PR #462, opened 2022-05-08
fix: Fix build error on Linux with gnustl

aa889788:
This error seems caused by gnustl vector, I do not know
how to fix it, so I use macro to change container to deque
under gnustl, use deque under other STL library will also
cause problem.


MistEO:
Thanks for your contribution! ✿✿ヽ(°▽°)ノ✿

PR #988, opened 2022-06-26
feat: add multiple timers support (up to 8)

GalvinGao:
This PR adds multiple timers support, in case users wish to set a more granular timer schedule.

In the future, we could use a more advanced cronjob-like scheduler, allowing users to add arbitrary triggers for their needs.

DD05FDBE1AB1E49DF1D16D5667435F27


MistEO:
数组呢呜呜

GalvinGao:
不会写呜呜

PR #1328, opened 2022-07-26
YoStarEN: fix stuck at reward claiming screen

AkarinVS:
This appears to be last missing template when claiming rewards.
However, I'm sure if EN requires the [Roguelike1GetDropSelectReward](https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/master/resource/template/Roguelike1GetDropSelectReward.png) template.


MistEO:
Thanks for your contribution!

PR #1774, opened 2022-09-06
Update 1.1-詳細な説明.md

wallsman:
Japanese text Tweaked（直訳っぽい日本語を調整）/  Unsupported add（未対応部分を注釈追記）


Tastror (approved):

MistEO:
Thanks for your contribution.

But if there are Japanese players using the CN client, they may mistakenly think that these features are not supported by the CN client after reading the readme. Just a guess.

wallsman:
For this reason, " JP" has been annotated, so players who can read Japanese should have no problem with this.

Furthermore, if the user base is at the level of playing the CN version, they will definitely have the skills to solve problems on their own.

Thank you for your concern.

PR #1821, opened 2022-09-12
Doc::Tweak JP text

wallsman:
JP text tweak（日本語の翻訳を若干修正）


Tastror:
The word「かんぺき」means being supported perfectly (do not need to change the config, etc.), compared to other supported client like Bluestack Hyper-V, whose configuration need to be type in manually.

Maybe we can add other words to indicate this?

wallsman:
If the word "完璧（かんぺき）" is used, it may also be required to address problems with the emulator (including improper specifications of the emulator or user).

If a supplement is to be added, it might be better to use a milder term such as "（開発の推奨）”.

Tastror:
I think「（開発の推奨）」is a good idea

Tastror:
By the way, you can edit all the files you want **before** handing in a pull request (then all the file changes will gather together in one pull request).

You can also continue your change in your own repository after handing in the pull request, the new commit will automatically add to your last pull request.

wallsman:
OK, I will be careful next time.

Tastror (approved):

PR #1827, opened 2022-09-12
docs: Update Chinese user manual

MistEO:


MistEO:
Could you help me to update the foreign language docs? @LYZhelloworld @Tastror @wallsman @nvkou

MistEO:
> No problem, but I can only play the JP version, is that ok?

Thanks! If what I wrote is hard to understand, you can also wait for several other collaborators to translate them into English or Japanese, then make them more authentic.

PR #1918, opened 2022-09-23
doc:Create 3.6-インフラスケジュール設定.md

wallsman:
CN base in Japanese, Maybe the translation is a little suspect.

https://github.com/wallsman/MaaAssistantArknights/blob/dev/docs/3.6-%E5%9F%BA%E5%BB%BA%E6%8E%92%E7%8F%AD%E5%8D%8F%E8%AE%AE.md


Tastror (approved):

docs/ja-jp/3.6-インフラスケジュール設定.md
```diff
@@ -0,0 +1,108 @@
...
+                                        // autofill=true との互換性はありません。空白でない場合、autofill は false である必要があります。
+                            "アステジーニ",
+                            "フィリオプシス",
+                            "サイレンス"
+                        ]
+                    }
+                ],
+                "meeting": [
+                    {
+                        "autofill": true // この施設はすべて autofill
```
MistEO, line 92: The Chinese here is "This room is autofill", I used DeepL to tell me that "All of these rooms are autofill" here in Japanese, this may not be true. @Tastror

Tastror: すべて（全部）是修饰 autofill 的，不是修饰 施設 的。施設 前面是 この（这个）。
所以不要用 DeepL 翻日语.jpg

wallsman: Change Text.
Change from whole sentences to single words.

//
どちらにせよ応接室は現状固定施設のため、”応接室すべて”でも問題はないようには思うのですが・・・
今後のアップデートで増えたりする可能性はあるので上記の方が良いかもしれませんね。

Either way, I don't see the problem with "この施設はすべて autofill" since reception rooms are currently fixed facilities...
However, it is possible that the number may increase in future updates, so the above may be a better choice.

docs/ja-jp/3.6-インフラスケジュール設定.md
```diff
@@ -0,0 +1,108 @@
...
+                    "23:59"
+                ],
+                [
+                    "00:00",
+                    "06:00"
+                ]
+            ],
+            "duration": 360,        // 作業時間(分) 予約項目, 未実装 将来的には、シフトを変更時間を告知させるポップアップ時間になるかも？ 自動的に変更される機能になる可能性もあり
+            "Fiammetta": {          // “フィアメッタ” どのオペレーターを配属するか，任意，記入しない場合は配属しない。
+                "target": "シャマレ",   // ターゲットオペレーター この記述例の場合フィアメッタとシャマレの体力交換スキーム
```
wallsman, line 29: I forgot to ask the question.
As for the target field, in OCR it is client language dependent, but which is the notation standard here?

docs/ja-jp/3.6-インフラスケジュール設定.md
Tastror: 気にしないで、これは DeepL 翻訳の問題です

docs/ja-jp/3.6-インフラスケジュール設定.md
Tastror: Maybe use this is OK.
What's the other possible notation standard?

wallsman: 下の方のOCR使用部分はクライアント依存であったため、こちらの部分はOCR使用なのか内部データ使用なのか判断ができないので質問させていただきました。
クライアント依存のOCR使用で問題ないのであればこのままPushしていただいて大丈夫だと思います。

I asked this question because I am unable to determine whether this part uses OCR or internal data, since the part below that uses OCR is client-dependent.
If there is no problem with client-dependent OCR use, I think it is fine for you to push as is.

Tastror: 谢谢啦~
不用担心OCR这部分的内容，这里的东西在之后OCR代码更改的时候会由我们相应修改的。现在直接写原名就是最好的写法。

Thank you~
Don't worry about the content of the OCR part, the things here will be modified by us when the OCR code changes later. Now writing the original name directly is the best way to write it.

docs/ja-jp/3.6-インフラスケジュール設定.md
MistEO: Users may refer here and then use `autofill` for other facilities that have multiple rooms. No problem with the new commit.

PR #2255, opened 2022-10-09
perf: Hide MAA from the Windows Task Switcher when it is minimized

gdlcf88:
image


MistEO:
Thanks for your contributions!

PR #3298, opened 2023-01-06
perf:improve the loading of config in rust bindings

KevinT3Hu:
Improve how CONFIG loads config file so that now it will create a default config file if not already present in the current directory.


MistEO:
Thx-ld

PR #3822, opened 2023-02-25
doc : Blue-ray items

wallsman:
Upadate text


liuyifan-eric:
I think the Bluray item is only available in JP so it is not necessary as of now?

wallsman:
I added this item because I thought it was necessary, but if don't need Close it plz.

Issue #4006, opened 2023-03-15
In roguelike mode, the bot stops after the first battle.


Eundong:
### Before raising the issue...

- [X] I understand that Issues are for feedback and problem solving, not trolling the comments section, and will provide as much information as possible to help solve the problem.
- [X] I filled in a short, clear title so that developers could quickly identify the general problem when going through the issue list. Instead of "Some suggestions", "Stuck", etc.
- [X] I have checked the latest beta update and there is no mention of the bug being fixed.

### Description

Client: Korea Server

The bot stops on the rewards screen after the first battle. It doesn't work since the latest patch. Everything else works fine except for roguelike mode, and changing emulators doesn't fix the bug.

### Log files

[asst.log](https://github.com/MaaAssistantArknights/MaaAssistantArknights/files/10978016/asst.log)
time : [2023-03-15 02:20:25.713]

### Screenshots

![스크린샷 2023-03-15 021907](https://user-images.githubusercontent.com/4812548/225264904-15184f09-21d9-493b-aab7-ef214feaaa2e.png)

### Anthing else?

Emulator : Nox or Bluestacks

PR #4093, opened 2023-03-27
fix: clientKR Roguelike@Dropsflag update

Eundong:
This flag was changed in a recent patch, as the result, the post-battle reward screen was not properly reconized


Eundong:
related issue is #4006

MistEO:
Thx-ld

PR #4724, opened 2023-05-10
fix(avatarcachemanager): avatar saved at wrong path in set_avatar()

lengyanyu258:
m_save_path is empty & ::load() never called when cache folder does not exist.


MistEO:
Thx-ld

PR #4894, opened 2023-05-23
Update for YostarKR

178619:
Makes Mizuki theme roguelike available for KR (initially) and updates localization


178619:
For now it barely makes 1st level farming method available. (Although it *can* reach 3rd and *probably* even 5th) I didn't have enough time to test it for so long therefore I focused on the farming method since it doesn't take too much time. I will come back to update it again some time later.

(You were also so fast on merging it by the way)

PR #4930, opened 2023-05-25
Localization fix for ko-KR

GyeRyak:
Social point shop default blacklist and whitelist not work properly due to the space, so removed it.
And slightly trimmed some other localizations.


liuyifan-eric:
> Social point shop default blacklist and whitelist not work properly due to the space

Can you specify this? I personally think it should be a bug

GyeRyak:
Yes, when the word containing whitespace like previous default "즉시 완료 허가권", OCR cannot detect it properly.
After I removed whitespaces, MAA works as intended.

> > Social point shop default blacklist and whitelist not work properly due to the space
> 
> Can you specify this? I personally think it should be a bug
> 
> Do you mean the items should have whitespace but the OCR cannot handle them?

MistEO:
I think this problem can be solved by adding `CreditShop-ProductName` - `ocrReplace` to YostarKR.

MistEO (approved):
LGTM. But another PR #4940 also changed a lot of translated text, maybe causing conflicts. We need to wait until #4940 is merged before merging this.

Thx-ld

PR #4945, opened 2023-05-27
修复EN服肉鸽不抓预备干员&添加百嘉鸿雪识别容错

Pitiedwzr:
maa识别预备干员时总是识别为“Reserve Operator  XXXXXX”缺少“-”，因此不抓预备干员


resource/global/YoStarEN/resource/tasks.json
```diff
@@ -2079,10 +2099,18 @@
...
+                "Pozemka",
+                "鸿雪"
             ],
             [
                 "Pozëmka",
                 "鸿雪"
+            ],
+			[
+                "Gavial the",
+                "百炼嘉维尔"
```
178619, line 2113: It may cause an issue that it converts `Gavial the Invincible` to `百炼嘉维尔 Invincible`. It might be better to put `Gavial the$` or merge those two into `Gavial the.*`

Pitiedwzr: Thank you for the correction, I'll also merge `Pozëmka` and `Pozemka` into `Poz.*mka`.

Constrat: Hey, just wanted to let you know that the problem is:

Before                 "Gavial the",  "百炼嘉维尔", there's already "Gavial", "嘉维尔". I went around and noticed that she's the only alter without the syntax "^Gavial$", Nearl, Skadi, Specter and the others have it. I changed it and now it works fine even without using "Gavial the.*".

I also wanted to thank you for the early implementation of the Mizuki Roguelike. And trying to do something by myself but I'm having really a hard time. Maybe you've already fixed this but, just wanted to let you know.

PR #4972, opened 2023-05-28
修复EN服肉鸽不放预备干员，不选非默认分队，添加复用安全屋文本

Pitiedwzr:
Fix YoStarEN roguelike does not deploy reserve operator and does not select non-default squad, add text of reuse safe house


Constrat:
Should also change `"Texas"` to "^Texas$" as she's an alter just like Gavial
Minor operators are also Hibiscus and Greyy.
In the latest alpha the tasks.json is still `"Hibiscus"` and `"Greyy"`
image

Pitiedwzr:
> Should also change `"Texas"` to `"^Texas$"` as she's an alter just like Gavial Minor operators are also Hibiscus and Greyy. In the latest alpha the tasks.json is still `"Hibiscus"` and `"Greyy"` image
> 
> The other alters are still CN only so it shouldn't be a problem, for now.

Thank you for reminding me, I forgot these alter operators, I’ll add them now.

Issue #4978, opened 2023-05-28
Mizuki Roguelike: The bot can't choose between two recruit permits


Constrat:
### Before raising the issue...

- [X] I understand that Issues are for feedback and problem solving, not trolling the comments section, and will provide as much information as possible to help solve the problem.
- [X] I filled in a short, clear title so that developers could quickly identify the general problem when going through the issue list. Instead of "Some suggestions", "Stuck", etc.
- [X] I have checked the latest beta update and there is no mention of the bug being fixed.

### Description

I've been using the 19.0 beta1 and beta2. When you finish the stage and the game gives you the option to choose between two recruits the bot holds gets stuck, doesn't choose any of them and drops the run, just to run again.

### Log files

[asst.zip](https://github.com/MaaAssistantArknights/MaaAssistantArknights/files/11585495/asst.zip)
16:06:29, 16:22:44, 16:53:43, 17:09:20, 19:37:14, 20:11:29, 20:27:57, 20:38:22, 20:52:24, 21:06:54. Here are the timestamps of the images in the `debug` `roguelike` folder.

### Screenshots

![2023-05-28_16-06-29-332_raw](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/755b620c-4c48-4433-a52a-796f833535bd)
16:06:29
![2023-05-28_21-06-54-263_raw](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/2e7c0cc0-2a79-4593-a370-32f60e7bf7a3)
21:06:54
![2023-05-28_16-22-44-259_raw](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/49de1992-5357-4c26-8996-41be864a208a)
16:22:44

Here are two examples of where it gets stuck. I have at least 10 other debug screenshots (that happened between this two) that MAA created automatically in the `debug` folder.

### Anthing else?

Another thing that I'm noticing is: in a lot of these screens you can notice that at the 3rd floor of the run I still have the default comp composed of the first 3 recruitments. Sometimes if lucky it will E2 my Gavial Alter (which is the only >3* operator, so the only upgradable one) I'm guessing that it skips every other recruitment option.

178619:
[...]

PR #4985, opened 2023-05-29
Update tasks.json (EN, #4978)

178619:


ABA2396:
Speaking of which, why are there two identical entries in the code block:
```
            [
                "Stainless™ Multifunctional Platform",
                "白铁™多功能平台"
            ],
            [
                "Stainless™ Multifunctional Platform",
                "白铁™多功能平台"
            ],
```
And it seems that the symbol '™' might not be recognized easily. How about replacing it with '.?' or '.*'?

Also, regarding the following entries:
```
            [
                "First-aid Kit",
                "急救包"
            ],
            [
                "Tide-Hunt Knight",
                "猎潮的骑士"
            ],
            [
                "8-Sided Dice",
                "8面骰子"
            ]
```
Would the dash '-' be correctly recognized?

178619:
>Speaking of which, why are there two identical entries in the code block: 

I think the Auto Update thing caused it. KR one had those, pretty sure JP one also has those duplicates.
>Would the dash '-' be correctly recognized?

I have no idea, can anyone who plays EN confirm this?

Constrat:
Honestly, I don't think that was an issue because during yesterday testing the only problem was the recruit... But maybe I wasn't lucky enough to get any of the items you listed above, that have a `-`.

Constrat (approved):
Works great!
https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/cf1b8e2e-b8b9-4123-bfe1-ffd1eeb30165

PR #4993, opened 2023-05-29
YostarEN Mr. Nothing, Shirayuki not recognized because of capitalization error / missing space in tasks.json

Constrat:
In game names are different from the tasks.json.
It looks like the names are from Aceship. It capitalizes the Y in ShiraYuki which is wrong in-game. Same thing with Mr. Nothing, it removes a ` `, whereas in game there is one.

I still can't understand what's wrong with Nine-Colored Deer.
image
The asst.log only shows
image
Sorry about double request but had a problem with the prior one.

_still getting used to GitHub_


resource/global/YoStarEN/resource/tasks.json
```diff
@@ -1285,7 +1285,7 @@
                 "蓝毒"
             ],
             [
-                "ShiraYuki",
+                "Shirayuki",
```
MistEO, line 1288:
```suggestion
                "Shira[Yy]uki",
```

resource/global/YoStarEN/resource/tasks.json
```diff
@@ -1561,7 +1561,7 @@
                 "霜华"
             ],
             [
-                "Mr.Nothing",
+                "Mr. Nothing",
```
MistEO, line 1564:
```suggestion
                "Mr.*Nothing",
```

MistEO:
For Nine-Colored Deer, I made a modification and moved some parameters to the json, and [2] should help solve this problem. You can try to modify this parameter in EN's tasks.json to see if it works, for example:

```json
        "specialParams": [
            160,
            4,
            6,
            0
        ]
```

PR #5042, opened 2023-06-03
Update YostarEN tasks.json for OperBox detection patch / suggestion

Constrat:
Found a way to make alters detectable. (kind of)

Ch'en the Holungday 假日威龙陈 and Gavial the Invincible 百炼嘉维尔 still can't be detected
Line 461 and line 716 [asst.log](https://github.com/MaaAssistantArknights/MaaAssistantArknights/files/11643351/asst.log)

258 / 260 operators are detected. (260 is the max number of ops in as of today). _**I'm open to other solutions**_ because honestly don't really like what I did to detected alters (Ch'En and Gavial are still not detected).

I feel like the OperBox detection should have it's own dedicated section in the tasks.json because for the alters, having double lines to make sure it works in every other detection I don't think is good.

Also Yostar really tried their best to fuck up how the names are rendered in the operator box, by giving alters stupid long names. Nearl Alter and still fit in one line (24 chars), Skadi Alter doesn't fit (26 chars) and Greyy Alter doesn't fit (25 chars).
I feel like the Oper Box task needs custom ROI values. Maybe not for all operators but only for certain ones. From my understanding it's not possible to add ROI to the ocrReplace because it works for every single char replacement.

### DIFF
--- a/resource/global/YoStarEN/resource/tasks.json
+++ b/resource/global/YoStarEN/resource/tasks.json
@@ -757,6 +757,10 @@
                 "幽灵鲨"
             ],
             [
+                ".*Radiant Knight",
+                "Nearl the Radiant Knight"
+            ],
+            [
                 "Nearl the Radiant Knight",
                 "耀骑士临光"
             ],
@@ -1189,6 +1193,10 @@
                 "远牙"
             ],
             [
+                ".*the Holungday",
+                "Ch'en the Holungday"
+            ],
+            [
                 "Ch'en the Holungday",
                 "假日威龙陈"
             ],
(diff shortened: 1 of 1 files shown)


MistEO:
This problem can be solved by adjusting the trim threshold, I am too busy to do it recently.

Constrat:
I tried searching a bit better in the CN tasks.json.
`    "OperBoxNameOCR": {
        "baseTask": "CharsNameOcrReplace",
        "text": [],
        "roi": [
            4,
            269,
            122,
            14
        ],
        "specialParams": [
            160,
            4,
            4,
            0
        ]
    },`
Could this be it? Just copying this in the EN one (without inheriting the CN) and changing the special Params? (From my understanding the ROI shouldn't be changed right?

Issue #5004, opened 2023-05-30
YostarEN: Nine-Colored Deer, Alter-Operators detection error.


Constrat:
### Before raising the issue...

- [x] I understand that Issues are for feedback and problem solving, not trolling the comments section, and will provide as much information as possible to help solve the problem.
- [X] I filled in a short, clear title so that developers could quickly identify the general problem when going through the issue list. Instead of "Some suggestions", "Stuck", etc.
- [X] I have checked the latest beta update and there is no mention of the bug being fixed.

### Description

Here from #4998 @MistEO  for Nine-Colored Deer. Since I'm opening this issue might as well mention the Alter-Operators problem.

### Log files

[asst.log](https://github.com/MaaAssistantArknights/MaaAssistantArknights/files/11600750/asst.log)
At line 2243 and 2244. For Nine-Colored Deer.

### Screenshots

![2023-05-30_15-55-50-878_raw](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/e44aaa2f-b501-4006-a2ad-c1b908c9e5d4)
![2023-05-30_15-55-30-510_raw](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/6e8fb675-7387-48ae-8c82-0084f2e8c349)
![2023-05-30_15-55-29-051_raw](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/467808cc-7375-4bbf-94ef-bb9f5b1c80ba)
![2023-05-30_15-55-28-194_raw](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/27c9418d-7946-48d1-ae9c-1692b77b1eaa)
![2023-05-30_15-55-49-338_raw](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/e2f57161-c111-4e62-aa02-7acf679627e6)
![2023-05-30_15-55-47-899_raw](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/afb793dd-5913-4657-9a50-be4ca2220222)

PR #5059, opened 2023-06-04
Update tasks.json for OperBoxNameOCR YostarEN

Constrat:
260/260 operators detected, finally fixed.

@MistEO the
`        "specialParams": [`
`            160,`
`            4,`
`            6,`
`            0`
`        ]`
from #5004 with [2] = 6 was a godbless, basically all the work that I had been doing this week has been rendered useless XD. I just tried tonight changing it and everything fixed itself. It took me a week because I had to understand a bit better the tasks.json and the MaaCore.

### ISSUE
The only that is left is Operator's names that have a new line in them, such as Skadi the Corrupting Heart and Greyy the Lightningbearer.

### DIFF
--- a/resource/global/YoStarEN/resource/tasks.json
+++ b/resource/global/YoStarEN/resource/tasks.json
@@ -2169,7 +2169,7 @@
                 "星源"
             ],
             [
-                ".*Lightnin.*bearer",
+                ".*Lightni.*bearer",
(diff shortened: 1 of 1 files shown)


MistEO:
Yes, I think this is the best solution. Thanks for your contribution!

PR #5081, opened 2023-06-06
Enhancement: Translated maa_tasks_schema.json

Constrat:
Used Google Translate to create a copy of the original one, just translated in english to get a better understanding of the tasks.json.


Constrat:
Have no idea what GitHub is doing. Wanted to make two different pr's and this guy tries to commit to the same pr

Issue #5060, opened 2023-06-04
MAA Does Not Recognize Adverse Environment Battle Start Button


SpinalWrithe:
### Before raising the issue...

- [X] I understand that Issues are for feedback and problem solving, not trolling the comments section, and will provide as much information as possible to help solve the problem.
- [X] I filled in a short, clear title so that developers could quickly identify the general problem when going through the issue list. Instead of "Some suggestions", "Stuck", etc.
- [X] I have checked the latest beta update and there is no mention of the bug being fixed.

### Description

When using MAA perform the fight task in Chapter 10 missions on the "Adverse Environment" difficulty MAA recognizes the initial "start battle" button, but when it gets to the squad selection screen with the "operation start" button it hangs for about 30 seconds. Then MAA sends back "Task: error Fight" and then stops the task.

### Log files

[asst.log](https://github.com/MaaAssistantArknights/MaaAssistantArknights/files/11646781/asst.log)

### Screenshots

![image](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/135544700/ffe81cdd-302a-4076-904d-487ee149e65d)
![image](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/135544700/eb62290b-5d17-492b-a0de-ed2737418ec5)

### Anthing else?

Emulator: Blues Stacks
OS: Windows 10
Client/Server Type: YoStarEN
MAA Version: v4.19.0

Constrat:
Fixed in #5020.

ABA2396:
@Constrat By the way, if it only involves modifying 'tasks.json', it can be achieved through modifications in [MaaRelease](https://github.com/MaaAssistantArknights/MaaRelease/tree/main/MaaAssistantArknights/api/resource/) for hot updates. Users just need to restart the client or wait for the automatic update at 4 or 16 o'clock to resolve it. However, it's important to note that updates that involve pictures or other files cannot be achieved through this method of hot updates.

Constrat:
> @Constrat By the way, if it only involves modifying 'tasks.json', it can be achieved through modifications in [MaaRelease](https://github.com/MaaAssistantArknights/MaaRelease/tree/main/MaaAssistantArknights/api/resource/) for hot updates. Users just need to restart the client or wait for the automatic update at 4 or 16 o'clock to resolve it. However, it's important to note that updates that involve pictures or other files cannot be achieved through this method of hot updates.

[...]

PR #5183, opened 2023-06-16
fix, style: Fight Task Adverse Start #5060

Constrat:
Added the StartButton3.png to CN/template and EN/template, modified CN tasks.json to implement new StartButton3 task.

From log:
asst::TemplResource::get_templ lazy load `StartButton3.png` match_templ | StartButton3.png score: `0.972579` rect: [ 1038, 371, 131, 272 ] roi: [ 1000, 350, 220, 340 ]


Constrat:
In the end I decided to got with the full copy of StartButton2 to StartButton3 and just change some bits.

Other options:
-  just changing the `"templThreshold":` to 0.70 (the log befoew this change shows that the score is 0.72) so 0.7 shoud work, This shouldn't create a lot more problems because the ROI is always in that specific place and the surrounding image is mostly just white.
- giving the task two templates to check
- Completely changing all of the fight tasks to implement the BattleStartAdverse.png and BattleStartNormal.png that already exist.

The solution that I chose is a bit long and not really made for performance or readibility of the code but should work and was easy to implement, just changing the threshold was easier for sure but I feel like that's just lazy,  giving a completely different image to check and hoping that the score matches.

resource/tasks.json
```diff
@@ -3081,6 +3084,20 @@
...
     },
+    "StartButton3": {
+        "action": "ClickSelf",
+        "preDelay": 1000,
+        "roi": [
+            1000,
+            350,
+            220,
+            340
+        ],
```
ABA2396, line 3095: Content like this, which is similar in nature, can actually be abbreviated as `"baseTask": "StartButton2"`. However, it's fine if you don't make the change; it doesn't really matter.

ABA2396 (approved):

resource/tasks.json
```diff
@@ -3136,9 +3109,22 @@
...
+        ]
+    },
+    "StartButton3TryAgain": {
+        "baseTask": "StartButton2TryAgain",
+        "template": "StartButton3.png",
+        "next": [
+            "StartButton3#next"
+        ]
+    },
     "StartButton3WaitTime": {
```
Constrat, line 3126: Tyvm for the heads up, one more question. how does the baseTask interpret the "next" inheritance. For example in this case can I just use `"baseTask": "StartButton2WaitTime",` and `next: ["StartButton3TryAgain"]` because it inherits the other options such as PRTS#Next, EndOfAction etc... Or I need to repeat all of the next like here.

resource/tasks.json
```diff
@@ -3136,9 +3109,22 @@
...
-        "algorithm": "JustReturn",
-        "postDelay": 40000,
+        "baseTask": "StartButton2WaitTime",
         "next": [
             "PRTS#next",
             "EndOfAction",
```
Constrat: Tyvm for the heads up, one more question. how does the baseTask interpret the "next" inheritance. For example in this case can I just use `"baseTask": "StartButton2WaitTime",` and `next: ["StartButton3TryAgain"]` because it inherits the other options such as PRTS#Next, EndOfAction etc... Or I need to repeat all of the next like here.

ABA2396: The "next" field needs to be written in its entirety. Whenever a display description is provided, it will directly override the previous configuration.

Constrat: Oh completely override, perfect than!
Thanks.

Constrat:
I honestly have no idea what GitHub is doing, I'm working on completely different commits and it pushed these commits in the wrong pull request, not allowing me to create another pull request... lmao

Constrat:
I'll create two new PRs, sorry for the trouble but Github is making me go crazy

Issue #5235, opened 2023-06-23
AutoSquad Copilot Operator's name Detection in YostarEN


Constrat:
### Before raising the issue...

- [X] I understand that Issues are for feedback and problem solving, not trolling the comments section, and will provide as much information as possible to help solve the problem.
- [X] I filled in a short, clear title so that developers could quickly identify the general problem when going through the issue list. Instead of "Some suggestions", "Stuck", etc.
- [X] I have checked the latest beta update and there is no mention of the bug being fixed.

### Description

The Auto Squad task uses the default CharsNameOcrReplace but (thanks Yostar) the way the names are handled is completely different. I've been testing this afternoon something a bit strange:

### First Issue:
Depending on the display resolution of the emulator "Skadi the Corrupting Heart" can either have a "\n" or not in the name
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">
<head>
<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/David/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/David/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
</head>
<body link="#0563C1" vlink="#954F72">

Resolution | "Operators" | "Squads"
-- | -- | --
720P | ![image](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/23c0317a-7bc3-41c0-a667-f1f7ff78af63) | ![image](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/6cb09727-996c-4afc-aa90-eafc09c6f39f)
1080P | ![image](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/5a645d11-4e4b-40bc-a675-28c0b46e1b4d) | ![image](https://github.com/MaaAssistantArknights/MaaAssistantArknights/assets/56174894/abfdcbd4-0bd5-4a31-88a0-de5c252919f3)
[...]

PR #5249, opened 2023-06-25
fix: Modified operators regex for Auto Squad and role button

Constrat:
From  #5235 
Most of the operators get detected, i'm noticing some issues with Ch'En the Holungday and Hibiscus the Purifier that don't get detected. Asst.log doesn't show anything of those two operators. Sometimes both Ch'en Alter and Hibiscus get detected, sometimes not. ~~I think I'm noticing a pattern: if they are on the top row of operators they detected, otherwise they don't.~~
Tried it again and it doesn't pick up Hibiscus alter even if it's on the top row

In this screen I tried getting the bot to pick Bagpipe first and then Ch'En alter
Screenshot_2023 06 25_14 45 13 534

14:44:56.994 timestamp for Ch'en undetected (BattleQuickFor
[asst.log](https://github.com/MaaAssistantArknights/MaaAssistantArknights/files/11859750/asst.log)
mation-OperNameFlag.png gets detected correctly)

Question:
in the EN tasks.json with some opeators I can use this syntax
            [
                ".*arl the Radiant Knight",
                "Nearl the Radiant Knight"
            ],
            [
                "Nearl the Radiant Knight",
                "耀骑士临光"
            ],

and it works as expected replacing the Word at every step.
But sometimes it doesn't work like here:
            [
                ".*cter the Unchained",
                "归溟幽灵鲨"
            ],
            [
                "Specter the Unchained",
                "归溟幽灵鲨"
            ],
I have to manually place the chinese characters on the first call of the operator.
Having the second ocrReplace is now useless because they always pass on the first one. Is this intended or am I missing something?


MistEO:
In the operator identification function, we use the role icon as the detection object. But in the formation function, we use the corner icon as the detection object, which is not good, because when the name is very long, the corner will be blocked and cannot be detected. We may change it to the same way as operator identification in the future.

MistEO:
We can still merge this PR for now, as a temporary solution. Thanks for your contribution!

Constrat:
No problem

> In the operator identification function, we use the role icon as the detection object. But in the formation function, we use the corner icon as the detection object, which is not good, because when the name is very long, the corner will be blocked and cannot be detected. We may change it to the same way as operator identification in the future.

But looking at the logs, the corner of the operator box gets identified correctly:
{ rect: [ 379, 603, 12, 20 ], score: 0.976320 }
The 379, 603 coordinate is the bottom row first box. But if you are telling me that the formation function is planned to go the same as the identificator I won't pry more. Thank you for the quick answer!

MistEO:
> I have to manually place the chinese characters on the first call of the operator.
Having the second ocrReplace is now useless because they always pass on the first one. Is this intended or am I missing something?

Oh I didn't notice that, 

- Only the first is needed, the second is redundant. 

	```json
	            [
	                ".*cter the Unchained",
	                "归溟幽灵鲨"
	            ],
	            [
	                "Specter the Unchained",
	                "归溟幽灵鲨"
	            ],
	```

-  Can be merged into one. 

	```json
	            [
	                ".*arl the Radiant Knight",
	                "Nearl the Radiant Knight"
	            ],
	            [
	                "Nearl the Radiant Knight",
	                "耀骑士临光"
	            ],
	```
	to
	```json
	            [
	                ".*arl the Radiant Knight",
	                "耀骑士临光"
	            ],
	```

Constrat:
But the thing is... In the first case with Nearl, it works correctly it double replaces first the full name and then in Chinese but with specter, when I tried it only replaces the first time, without replacing the full name to Chinese. It seemed a little strange that in some cases it worked and in others it didn't. With Nearl and all the other operators I prefer having the double replace because In case the new one doesn't work, the other one works as expected in all the other game modes.
ALSO... While letting MAA do 10-16 I noticed that with Nearl alter new skin, che it tries in fight to detect the name it misses the last T in radiant kighT so I'll have to edit that out

Issue #4505, opened 2023-04-27
边玩游戏变挂MAA经常断连


chanaaifu:
### 在提问之前...

- [X] 我理解 Issue 是用于反馈和解决问题的，而非吐槽评论区，将尽可能提供更多信息帮助问题解决
- [X] 我填写了简短且清晰明确的标题，以便开发者在翻阅 issue 列表时能快速确定大致问题。而不是“一个建议”、“卡住了”等
- [X] 我已查看最新测试版本的更新内容，并未提及该 bug 已被修复的情况

### 问题描述

有时候玩游戏能完整挂完，有时候玩一半切出来一看就断连了，然后重新开MAA和重新开模拟器都不管用，模拟器用的蓝叠

### 日志文件

[2023-04-27 10:13:06.497][INF][Px75e4][Tx5758] Assistant::append_callback | ConnectionInfo {"uuid":"","details":{"adb":"C:\\Program Files\\BlueStacks_nxt_cn\\HD-Adb.exe","address":"127.0.0.1:5555","config":"BlueStacks"},"what":"ConnectFailed","why":"Connection command failed to exec"}
[2023-04-27 10:13:06.497][TRC][Px75e4][Tx5758] asst::AdbController::connect | leave, 108 ms
[2023-04-27 10:13:06.497][TRC][Px75e4][Tx5758] asst::MinitouchController::connect | leave, 108 ms
[2023-04-27 10:13:06.497][TRC][Px75e4][Tx5758] asst::MinitouchController::~MinitouchController | enter
[2023-04-27 10:13:06.497][TRC][Px75e4][Tx5758] asst::MinitouchController::~MinitouchController | leave, 0 ms
[2023-04-27 10:13:06.497][TRC][Px75e4][Tx5758] asst::AdbController::~AdbController | enter
[2023-04-27 10:13:06.497][TRC][Px75e4][Tx5758] asst::AdbController::~AdbController | leave, 0 ms
[2023-04-27 10:13:06.497][ERR][Px75e4][Tx5758] connect failed
[2023-04-27 10:13:06.497][TRC][Px75e4][Tx5758] asst::Controller::connect | leave, 108 ms
[2023-04-27 10:13:06.497][TRC][Px75e4][Tx5758] asst::Assistant::ctrl_connect | leave, 108 ms
[2023-04-27 10:13:06.497][TRC][Px75e4][Tx5758] asst::Assistant::connect | leave, 109 ms
[...]

PR #5666, opened 2023-07-27
Implement ADB hard restart.

AsterNighT:
### What does this PR do?
1. Implement ADB hard restart for maa. Now when used together with accelerators (like #4505), adb occasionally loses connection to the emulator. In this case, invoking `adb kill-server` and `adb start-server` will not solve the problem. But from my experience killing and restart ADB process itself does.
2. Move some of the logs (specifically the ones of restarting emulator and ADB) after the option checks. Now even you do not tick "连接失败后尝试重启 ADB", maa still print a line "正在尝试重启 ADB Server" though it actually does nothing.

### What is needed for this PR to be considered complete?
- [x] Tests, should be done later today.
- [x] Localizations. I know little about Japanese, Korean or Traditional Chinese. It will be great if someone could help me filling these blank or I would have to go to google translation.

### Concerns
In this PR maa is killing process possibly not run by itself. It could cause potential harm. I know not much about Windows permission management but in my tests maa could kill the adb process even when adb is launched by the administrator while maa is not.

Update: I can trigger this bug by connecting and disconnecting from the accelerator while maa is running. A further investigation may be conducted to find out the exact cause of it.


ABA2396 (approved):

AsterNighT:
Done. PTAL @MistEO @ABA2396

src/MaaWpfGui/ViewModels/UI/SettingsViewModel.cs
```diff
@@ -775,6 +776,49 @@ public void RestartADB()
...
+                            {
+                                Process = p,
+                                Path = (string)mo["ExecutablePath"],
+                            };
+                foreach (var item in query)
+                {
+                    if (item.Path == adbPath)
+                    {
+                        item.Process.Kill();
+                        item.Process.WaitForExit();
```
ABA2396, line 816: 这里得拿try catch包一下，有些模拟器的adb会是管理员权限

AsterNighT: 我拿管理员权限自己起了一个，但好像也能成功地kill掉，虽然我也觉得不应该行。

PR #5944, opened 2023-08-14
feat: merge and re-structure stage.json

### LINKED ISSUE #5881: Can't report Event Stages to Penguin in EN and JP
### Before raising the issue...

- [X] I understand that Issues are for feedback and problem solving, not trolling the comments section, and will provide as much information as possible to help solve the problem.
- [X] I filled in a short, clear title so that developers could quickly identify the general problem when going through the issue list. Instead of "Some suggestions", "Stuck", etc.
- [X] I have checked the latest beta update and there is no mention of the bug being fixed.

### Description

Title is pretty self explanatory: can't report `WB-7` drops to Penguin Statistics.
More generically I've noticed this issue with all Event Stages.
Normal stages don't seem to have this issue (For example 1-7)

### Log files

EVENT STAGE WB-7
Extract of the Report Task from the logs.
The "why" is UnknownStage, when diving the in log `"stage":{"stageCode":"WB-7","stageId":""}`
`stageId` is empty. 
[asst.log](https://github.com/MaaAssistantArknights/MaaAssistantArknights/files/12289421/asst.log)

NORMAL STAGE 1-7
`"stage":{"stageCode":"1-7","stageId":"main_01-07"}` stageId gets correctly matched.
[asst.log](https://github.com/MaaAssistantArknights/MaaAssistantArknights/files/12289703/asst.log)

### Screenshots

Screenshot_2023 08 08_10 21 17 114
image
### **――――――――――――――――――――――――――――――――――――――――**
Screenshot_2023 08 08_11 16 47 078
image
### Anthing else?

image
image
image
image

MistEO:
fix #5881


Constrat:
Testing this won't be easy as we don't have any non-rerun event opened.
Invitation to Wine was working properly (it's a rerun) apart front eh furniture recognition/report @wallsman was having issues with it. I still haven't found furniture's so I can't be sure

wallsman:
In fact, the drop rate of furniture is quite low and I have only been able to check it in real time, roughly 15 times in almost 1000 laps. I have only logs so I have not been able to take screenshots.

Constrat:
@MistEO Looking at the stages.json you've decided to discard all of the i18n, clear timings while leaving only the stageId (which is the most important part).

Is this what you were talking about the "legacy" file vs the current one?

MistEO:
> @MistEO Looking at the stages.json you've decided to discard all of the i18n, clear timings while leaving only the stageId (which is the most important part).
> 
> Is this what you were talking about the "legacy" file vs the current one?

Not only that, I merged the `dropInfos` of multiple servers together. Now this json contains drop info for all servers (instead of CN only)

Constrat:
> > @MistEO Looking at the stages.json you've decided to discard all of the i18n, clear timings while leaving only the stageId (which is the most important part).
> > Is this what you were talking about the "legacy" file vs the current one?
> 
> Not only that, I merged the `dropInfos` of multiple servers together. Now this json contains drop info for all servers (instead of CN only)

So no more differentiation between servers because we said before that MAA will NOT report stages that are not open right?

MistEO:
Yes, it contains all the stages that are opening.

PR #5948, opened 2023-08-14
refactoring YostarEN/resource/tasks.json

Constrat:
Removed Award (was a duplicate) + I.S. trader shopping regex + general refactoring


MistEO:
Since the auto-updater needs to update these files from time to time, we have no way to maintain the order of the fields, they will be automatically changed to alphabetical order in the next update. ;(

Constrat:
Does the auto updater also change the order inside the task?
Like "fullmatch" "text" "roi" etc?

MistEO:
> Does the auto updater also change the order inside the task?
> Like "fullmatch" "text" "roi" etc?

Yes, by definition, JSON object is unordered.

MistEO:
Some json parsing libraries implement ordered objects, but this is not required by the json standard. Sadly meojson doesn't implement it.

Constrat:
I said it wrongly: I meant inside a random task like this: 
It it-s like this will it order it?
```
    "SSSDropRecruitmentOCR": {
        "roi": [
                    0,
                    352,
                    1280,
                    263
                ],
        "postDelay": 1000,
        "action": "DoNothing",
        "algorithm": "OcrDetect",
        "cache": false,
        "baseTask": "CharsNameOcrReplace",
        "text": [
        ]
    }
```
    "SSSDropRecruitmentOCR": {
        "action": "DoNothing",
        "algorithm": "OcrDetect",
        "baseTask": "CharsNameOcrReplace",
        "cache": false,
        "postDelay": 1000,
        "roi": [
            0,
            352,
            1280,
            263
        ],
        "text": [
        ]
    }
```
Like this?

**Comment by MistEO** [comment]
Yes, I mean, it is reasonable for json objects to be changed to any order, and meojson will change them to alphabetical order. (Compared to the order of input, the alphabetical order is also a kind of disorder)

**Comment by Constrat** [comment]
> Yes, I mean, it is reasonable for json objects to be changed to any order, and meojson will change them to alphabetical order. (Compared to the order of input, the alphabetical order is also a kind of disorder)

Yeah when working on extended lines that are ordered logically, and getting is ordered alphabetically is a frustrating 😆

**Comment by MistEO** [comment]
Alphabetical ordering is not a good idea, as CN's tasks.json is all manually updated without auto-updater, so it doesn't have this problem. A better solution is to separate the part that needs to be automatically updated as another file, but it seems to require more modification.

**Comment by Constrat** [comment]
I think that doubling the jsons makes eveyrthing a bit messy because then you need to edit two files to get something to work. Is there no way to order it like the CN? Parsing the entire resource/tasks.json and then ordering the overseas ones like the CN? Missing tasks would be ignored.

**Comment by MistEO** [comment]
Sounds good, we need to add some code in the auto-updater to do the sorting, but it's not easy to implement in c++. Maybe we could have a python or js script that automatically runs after each run of the auto-updater to sort.

**Comment by Constrat** [comment]
Having for c++ for that would be a bit complicated and excessive, I think scripting languages would definitely be the way to go.

**Comment by Constrat** [comment]
@MistEO Ok I'm finished with the python script... Just a small issue is it normal that there's a task in the Korean, but not in the CN?
```
    "Mizuki@Roguelike@FromIntegratedStrategies": {
        "text": [
            "테마진입"
        ]
    },
```
```
    "Mizuki@Roguelike@SelectTheme": {
        "baseTask": "Roguelike@SelectTheme",
        "text": [
            "水月"
        ],
        "next": [
            "Mizuki@Roguelike@FromIntegratedStrategies"
        ]
    },
```
Apart from this everything works and gets ordered like the CN

**Comment by MistEO** [comment]
It may be that the CN was deleted, but the KR one was not deleted. XD

**Comment by Constrat** [comment]
And this never crashed anyones MAA? That's insane xD

**Comment by Constrat** [comment]
I can remove the Phantom and Mizuki one right? As they are child tasks

**Comment by MistEO** [comment]
@ABA2396 _(:з」∠)_

**Comment by Constrat** [comment]
Wait why did it try to check a full build?

**Comment by Constrat** [comment]
Ok @MistEO I've placed it in /MaaAssistantArknights/tools/Tasks_Sorter.py

**Comment by MistEO** [comment]
Thanks! you can add a step after https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/.github/workflows/res-update-game.yml#L31 to make it run automatically!

**Comment by Constrat** [comment]
> Thanks! you can add a step after https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/.github/workflows/res-update-game.yml#L31 to make it run automatically!

Gotchu! I'll try to normalize the path structuring

**Comment by Constrat** [comment]
@MistEO Quick question. When running actions the "prints" from python do you know where they go?
Because I have a choice to make.
Use "append" or use "insert". If I use insert I could add a small statistic which is: how many tasks are there in every tasks.json.
What do you think?

**Comment by 178619** [comment]
```py
import json
from pathlib import Path
from collections import OrderedDict

path = str(Path(__file__).parents[2]) + '/'

servers = {
    'EN': 'YoStarEN',
    'JP': 'YoStarJP',
    'KR': 'YoStarKR',
    'TW': 'txwy'
}

with open(path+'resource/tasks.json', 'r', encoding='UTF-8') as file:
    order = list(json.load(file).keys())
    print('CN', len(order))

for server, name in servers.items():
    with open(path+f'resource/global/{name}/resource/tasks.json', 'r', encoding='UTF-8') as file:
        tasks = json.load(file)
    tasks = OrderedDict(sorted(tasks.items(), key=lambda item:order.index(item[0]) if item[0] in order else -1))
    with open(path+f'resource/global/{name}/resource/tasks.json', 'w', encoding='UTF-8') as file:
        json.dump(tasks, file, indent=4, ensure_ascii=False)
    print(server, len(tasks))
```
```ls
CN 1278
EN 103
JP 54
KR 71
TW 71
```
I believe this is much simpler and does what you want to do...

Constrat:
Well it does. It was my first time using python so I tried my best XD... 8 hours went down the drain.

Constrat:
I see whatt you did. I had the idea at first to try and move everything using a "server" variable, but coming from a little programming in C and Java I basically had no idea how to implement it so I just went with "stock" coding.
Yours very pretty. Instead of having the 4 different server just using a variable's very cool!

178619:
> Now I see the issue: it's basically duped:
> <details>
>
> ```
>     "Roguelike@FromIntegratedStrategies": {
>         "text": [
>             "테마진입"
>         ]
>     }
> 
>     "Phantom@Roguelike@FromIntegratedStrategies": {
>         "text": [
>             "테마진입"
>         ]
>     }
> 
>     "Mizuki@Roguelike@FromIntegratedStrategies": {
>         "text": [
>             "테마진입"
>         ]
>     },
> ```
</details>

It's just my dumb mistake. I should've only put just one `Roguelike@FromIntegratedStrategies` when the task was added to CN but I didn't look into it properly and added those two instead. It should be safe to remove them.

178619:
I am pretty sure you can just use `print()`. Honestly I know nothing about how GitHub actions work even, but the logs do display `print()`, I could notice it by looking into [maadeps-download.py](https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/maadeps-download.py) and [the logs](https://github.com/MaaAssistantArknights/MaaAssistantArknights/commit/471ee5618729b2539405dcaca7ac510fd7db1e30/checks/15913595467/logs)...

> I see whatt you did. I had the idea at first to try and move everything using a "server" variable, but coming from a little programming in C and Java I basically had no idea how to implement it so I just went with "stock" coding.
Yours very pretty. Instead of having the 4 different server just using a variable's very cool!

Thanks, I am no good at programming but I've learned python a bit from doing my personal projects so I figured I could lend a hand...

resource/global/YoStarKR/resource/tasks.json
```diff
@@ -5367,7 +5357,8 @@
             352,
             1280,
             263
-        ]
+        ],
+        "text": []
```
MistEO, line 5361: This doesn't seem right...

MistEO: This `text` was the same as `CharsNameOcrReplace` before, but now it is empty.

Constrat:
@178619 I've completely forgot about this but... There was already a python script for sorting the global like the ZH... 
[tools/OverseasClients/SortJsonByZHServerOrder.py ](https://github.com/MaaAssistantArknights/MaaAssistantArknights/blame/dev/tools/OverseasClients/SortJsonByZHServerOrder.py)...Which you also modified! XD

PR #5977, opened 2023-08-17
fix: add ocr replace for YostarJP

jie17:
OCR 识别错误导致进不了商店

### DIFF
--- a/resource/global/YoStarJP/resource/tasks.json
+++ b/resource/global/YoStarJP/resource/tasks.json
@@ -2208,6 +2208,12 @@
     "Mall": {
         "text": [
             "買部"
+        ],
+        "ocrReplace": [
+            [
+                "買瑠",
+                "買部"
+            ]
         ]
     },
     "CreditStoreOcr": {


jie17:
Sure. This is the screenshot.
image
And the log

```
[2023-08-17 22:16:15.479][TRC][Px7f58][Tx7920] asst::PipelineAnalyzer::analyze Mall
[2023-08-17 22:16:15.494][TRC][Px7f58][Tx7920] asst::WordOcr [{ 購買瑠: [ 38, 29, 114, 51 ], score: 0.751537 }] by OCR Pipeline , cost 14 ms
```

Constrat:
If you could give me a full screenshot of the entire screen directly from the launcher I can try to change the ROI instead of changing the text as to also optimize the performance. (remember to cover name and UID)

MistEO:
For recognition errors, adding replacement should be a better choice. Others may wonder, what is "買瑠", but if it is in the replacement, we can easily see that this is a recognition error. Of course, it would be better if it can be solved by modifying the ROI.

Constrat:
> @Constrat For recognition errors, adding replacement should be a better choice. Others may wonder, what is "買瑠", but if it is in the replacement, we can easily see that this is a recognition error. Of course, it would be better if it can be solved by modifying the ROI.

Oh ok. I wasn't really sure about the double "text", that's why I wanted to see if maybe customizing decreasing the ROI works. But I think that the issue is: that it reads the shopping cart behind the last kanji so it has some weird lines

Constrat:
Ok for now let's go with this.

PR #5978, opened 2023-08-17
MAA doesn't click the trust notification in ZH_CN base

Constrat:
Could be similar to #292 @MistEO ?
**Log:**
```
asst::PipelineAnalyzer::analyze InfrastReward
asst::OcrPack::check_and_load | enter
det 1 rec 1 ocr 1
asst::OcrPack::check_and_load | leave, 177 ms
asst::WordOcr [{ 43点击全部收取: [ 205, 53, 116, 22 ], score: 0.973224 }, { 待办事项: [ 54, 83, 90, 23 ], score: 0.995768 }, { 干员信赖: [ 216, 80, 94, 29 ], score: 0.997068 }, { nn: [ 340, 85, 38, 10 ], score: 0.637426 }, { 线索搜集: [ 389, 83, 91, 23 ], score: 0.999981 }] by OCR Pipeline , cost 71 ms
Proced []
```
**Raw:**
image
### Potential Fix?
I don't want to directly commit for ZH_CN changes as I have a basic knowledge of the official server that's why I'm sking here also, it seems strange?
I've tested with `干员信赖` and now it works properly. Did they update this? From `员信赖` to `干员信赖`?
Or it worked before with the 4 characters one while testing the 3 one?


MistEO:
Which issue is this?

Constrat:
No issue. I was just using MAA for the CN account and noticed it was getting stuck in the operator trust notification (not max trust). I've checked the log and noticed it wasn't getting identified correctly

Constrat:
I don't understand something. When WordOcr detects something like `干员信赖`. When full match is false (default). Should `员信赖` work? Because I noticed in the SLChapterToSL WordOcr detects more stuff (incorrect) but part of the string is right and so it goes on without error.
In the base trust it doesn't. I had to manually extend the string

Constrat:
Ok I've tried it again and nothing happened... But looking at the log it never got detected, but it got clicked... Strange.

PR #6009, opened 2023-08-20
fix: all operators are marked as new for overseas clients (Toolbox-Recruitment)

178619:
Makes Maa use IDs for recruitment instead of names, so it can properly show potential data for overseas clients.

<details><summary>Screenshot (Before)</summary>

image
</details>

<details><summary>Screenshot (After)</summary>

image
</details>

### DIFF
--- a/src/MaaCore/Config/Miscellaneous/RecruitConfig.cpp
+++ b/src/MaaCore/Config/Miscellaneous/RecruitConfig.cpp
@@ -24,6 +24,7 @@ bool asst::RecruitConfig::parse(const json::value& json)
     for (const json::value& oper : json.at("operators").as_array()) {
         Recruitment oper_temp;
         oper_temp.name = oper.at("name").as_string();
+        oper_temp.id = oper.at("id").as_string();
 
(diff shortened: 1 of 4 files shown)


MistEO:
For users who have local cache before (operator recognition result), if they do not update the cache, the result will be wrong. Maybe we need to add some tips, or find a way to be compatible with the previous local cache.

MistEO:
Forget it, just mention it in the release notes. XD

Constrat:
> Forget it, just mention it in the release notes. XD

Yeah, people are going to complain that it's not working either way...

178619:
> For users who have local cache before (operator recognition result), if they do not update the cache, the result will be wrong. Maybe we need to add some tips, or find a way to be compatible with the previous local cache.

Actually, the `Configurations/Default/OperBox.Data` field in `gui.json` already contains IDs. I deleted the file then tried copying it from v4.21.0, it worked well.

Constrat:
nvm I think I found it.

Constrat:
Wait why are there three "OperBox.Data" keys in the Json?

PR #6022, opened 2023-08-21
fix: 修复干员识别错误

Black1312:
修复大批量干员识别多出下划线导致的自动编队错误


Constrat:
@MistEO This was caused by the algorithm change I've requested right?
I'll take a look at the EN

MistEO:
@Constrat This is to fix 8ca76de6f89a88ea9a817f9b30644315f5df0e16, it only affects the recognition of combat quick formation, and has nothing to do with operator box.

Constrat:
Wait I mean is: it has the same issues as it had the operbox.

With the "new" algorithm for battleformation, the Rhodes Island logo is in the way. In OperBox it works properly because it has the specialParams, am I wrong?

MistEO:
The problem with the Rhodes Island logo has always been there and I haven't changed it. If we didn't catch it before, it's probably because we rarely form 3-star Operators.

Constrat:
Oh I thought the specialParams where there for that reason... Maybe I never noticed either XD, my bad.
I'm going crazy for EN names, as they have more issues than anything else. I'm on the verge of dropping everything...

Constrat:
Quick question. Does the tasks.json of the official have options for operator `W`?
I don't know why but out of nowhere now it gives issues to EN. It gets detected to `w` so I just ocrreplaced it to `W`.
I'm guessing this doesn't happen in official right?

MistEO:
It seems that this problem does not exist in CN

Constrat:
Ok huge correction: I don't know how / what I did. There's no more problem with 3* operators... dafuq?
```
asst::WordOcr [{ Midnight: [ 0, 0, 64, 26 ], score: 0.957674 }] by OCR Rec , cost 8 ms
Proced [{ 月见夜: [ 918, 312, 64, 26 ], score: 0.957674 }]
asst::WordOcr [{ Melantha: [ 0, 0, 68, 26 ], score: 0.998704 }] by OCR Rec , cost 7 ms
Proced [{ 玫兰莎: [ 769, 590, 68, 26 ], score: 0.998704 }]
asst::WordOcr [{ Popukar: [ 0, 0, 62, 26 ], score: 0.999673 }] by OCR Rec , cost 8 ms
Proced [{ 泡普卡: [ 920, 590, 62, 26 ], score: 0.999673 }]
```



> It seems that this problem does not exist in CN

Ok, no problem, it's just replacing. (To be safe I'm adding `^w$`

PR #6028, opened 2023-08-21
Tweaked the regex for the BattleQuickFormation AutoSquad new algorithm

Constrat:
Mainly EN changes (I've tried and tested backward - compatibility. Everthing else should work fine... Hopefully)

I've removed `^w$` from txwy and EN and added it to the CN because I'm feeling it's an issue in all overseas clients.
This way it gets used by every server.

If it arises issues: no problem I can just add it back to EN and txwy


MistEO (approved):

Constrat:
I've stolen an account from my friend XD. He has every operator. I forgot about Greyy Alter which has a huge ass name. It's going to be a pain in the ass...

Constrat:
Adding patches over patches at this point... I feel like every emulator is going to render the name in a different way and at the end of it, it'll work only on my machine... Well if it doesn't work for the others they'll open an issue, there's not much else I can do.

`Ss?k?ad?i ?the ?Co?r*m?u?pt?i?ng.*` ... this looks ugly but it works, kinda

Constrat:
Ok with this one it should be done... (Hopefully) (Very Hopefully)

PR #6042, opened 2023-08-22
[JP] 春分 Update operators.md

wallsman:
[JP] 春分
ウインドフリット / ウィンドチャイム
The size of the "イ" in the name section is different, so it might be OCR caught?


Constrat:
> The size of the "イ" in the name section is different, so it might be OCR caught?

Isn't here ocr_config with the equivalance class between `イ` `ィ`?

Constrat:
@178619 already merged it so... Let's just deal with it. If issues arise we just need to ocrreplace it. Nothing game breaking (hopefully)

Constrat:
Closing, as all the proposed changes have been committed by @178619 also with KR support

178619:
This doc file does not change anything for OCR, you would have to edit tasks.json not this file. And yes, ocr_config exists so it won't matter anyway.

By the way I made a short script to automatically update this file, I think I will run it time to time...

PR #6228, opened 2023-09-05
[JP] readme Update test / Japanese font readability

wallsman:
The current default font specified is a Chinese-based font, which is very difficult to read, so a Japanese-based font was mandatorily specified.

現在デフォルトで指定されているフォントが中国語ベースのフォントで非常に読みづらくなっているので、日本語ベースフォントを強制指定した。

### DIFF
--- a/docs/ja-jp/readme.md
+++ b/docs/ja-jp/readme.md
@@ -1,4 +1,4 @@
-//Not Chinese Base font use Japanese(Better readability)
+<!-- 中華フォント優先のため、日本語フォント指定 / Not Chinese Base font use Japanese(Better readability) -->
 <font face="Helvetica Neue,Arial,Hiragino Kaku Gothic ProN,Hiragino Sans,Meiryo,sans-serif">
     
 <div align="center">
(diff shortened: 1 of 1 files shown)


wallsman:
If this is allowed, it will be applied to other pages as well.

MistEO:
image

This comment is displayed.

ABA2396:
```
<!-- 注釈此處 -->
```

wallsman:
Yeah, An embarrassing mistake has been corrected.

MistEO:
There seems to be no css file, I don't know, I don't know much about the front end.

@SherkeyXD Please help answer.

SherkeyXD:
> By the way, where is the CSS file for the page? It would be faster if you let us specify it there.

Our document is built under [Vuepress](Vuepress) using [Vuepress Theme Hope](https://github.com/vuepress-theme-hope/vuepress-theme-hope), so the css file is generated during the building process.
Unfortunately, it seems that this theme doesn't provide language-specific font settings (see [document](https://theme-hope.vuejs.press/cookbook/customize/font.html) ).

PR #6360, opened 2023-09-14
Add collab operator & fix some operators OCR

zewoosJ:
In KR, There was a problem that certain separation operators were not recognized by OCR, so we modified it to work by not recognizing existing operators again. like the last episode of "knight (나이트)".
The following operators are not recognizing anymore. Yato, Noir Corne.
Because alternative operators are more likely to be used in the R.I.I.C


MistEO (commented):
Is it incorrect that the original Yato and Noir Corne will no longer be recognized?

zewoosJ:
No, it was meant to be. OCR tends to recognize the operator name without the first letter, which causes overlapping operators (e.g., Justice Knight, Black Knight, Nearl the Radiant Knight) to collide with each other, causing only one common or no recognition (probably with the added Favorites feature, so I guess it's the most often used RIIC).
Therefore, if the names overlap, first of all, modifying them according to the most frequently used situations in RIIC was a fragmentary but not unreasonable way for users to use them right away, and this is how we did it.

Constrat:
Could you send some screens? Because with YostarEN I had some similar troubles because the names were too long 
This issues disappeared after we changed the main OCR name algorithm, which allowed for a bigger ROI. The only issue left was names that had a \n and went to a new line.
Unfortunately I haven't found a final fix for that. The "only" solution I came up with is this for Skadi Alter:
`Ss?k?ad?i ?the ?Co?r*m?u?pt?i?ng.*` it's ugly, but it does it's just. IMPORTANT: you should try to see how different resolutions affect the naming of the operators.
ref:https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues/5235

zewoosJ:
Oh... I'm in the military right now, so I did the modification once a month outside, so there's no way to access the program for a month...
Usually, I'm satisfied that other acquaintances are automatically running, so I can't afford to analyze the log and take screenshots right away.
I'll try to find people around me who can help me indirectly on this matter.

zewoosJ:
Did you mean the in-game operators screenshot?

Constrat:
Yeah because it might be similar to what happened in EN, fixing the names when they are in two different lines is a pain in the a**

MistEO:
The main problem is operator box recognition, this modification will tell all players that they have Kirin R Yato, and the information such as level is also wrong.

zewoosJ:
Here's a capture of all the operators in KR. I'm not sure if the picture will be uploaded properly. The last page is the offers that were mostly recognition errors.
Screenshot_20230914_202933
Screenshot_20230914_202938
Screenshot_20230914_202946
Screenshot_20230914_202953
Screenshot_20230914_203004
Screenshot_20230914_203010
Screenshot_20230914_203027
Screenshot_20230914_203032
Screenshot_20230914_203039
Screenshot_20230914_203145

Screenshot_20230914_203154
Screenshot_20230914_203159
Screenshot_20230914_203205
Screenshot_20230914_203219
Screenshot_20230914_203325

Constrat:
I can understand why Nearl alter and Ratapos could have issues, but with Kirin everything seems fine
@MistEO unless they don't have the Latin alphabet letters R and S?

zewoosJ:
As far as I remember, based on spacing, the first word was completely decimated and recognized, and both of them recognized it as the same letter.

wallsman:
Ah... The JP version actually has the same problem, Penance and Bena, two of the new additions, are experiencing the same problem, so I will refer you to the fix.

MistEO:
Converting all Yotos to Kirin R Yotos may be useful in the Copilot (auto-battle) function since the original Yotos are not commonly used. 

However, it will cause operator box recognition errors, and it will also cause the original Yoto to be recruited and frequently used in automatic IS mode (because Kirin R Yoto has a high priority). Overall, I don't think this is a good idea.

I guess we can decide whether to do the conversion by limiting the length of the word, like `.{3,}Yoto` ? (I haven't tested it, just an idea)

MistEO:
> I can understand why Nearl alter and Ratapos could have issues, but with Kirin everything seems fine
> @MistEO unless they don't have the Latin alphabet letters R and S?

Yes, we need the logs to know how to fix it.

Constrat:
> I guess we can decide whether to do the conversion by limiting the length of the word, like `.{3,}Yoto` ? (I haven't tested it, just an idea)

Oh good idea, only thing to worry about is OCR picking up something different when looking for Yato (non alter)

Constrat:
I've checked `keys.txt` for KR paddleocr and it has all the Latin chars, so this is not the issue.
Having logs would definitely speed this all up. @178619 have you found this issue on your end?

178619:
Sorry, I am unable to look into this matter for now. Well, I've got personal issues. However I can tell what I found or experienced much earlier.

> In KR, There was a problem that certain separation operators were not recognized by OCR, so we modified it to work by not recognizing existing operators again. like the last episode of "knight (나이트)".
> Therefore, if the names overlap, first of all, modifying them according to the most frequently used situations in RIIC was a fragmentary but not unreasonable way for users to use them right away, and this is how we did it.

This was undone by me already long ago. It had to be done as it was clear that it would cause false detection. This must be avoided even if avoiding it means Maa will be unable to detect some operators. Please don't try to 'fix' it in this way. It doesn't even work properly for RIIC as Maa will put an operator when '나이트' is detected even if that '나이트' isn't the one you intended to deploy. For other functions, MistEO already explained why it is a bad idea. This shall never be the way we fix it. Period.

Of course, although this method must be avoided, it doesn't change that there's an OCR issue we have to fix.

> I guess we can decide whether to do the conversion by limiting the length of the word, like .{3,}Yoto ? (I haven't tested it, just an idea)

This may work, but we still have to test it. The main issue is that it sometimes just slices the names when there's a whitespace. For most cases they are simply ignored or represented as some chracters (normal space, nbsp, a dot, etc.) so it is easy to be dealt with by adding ocrReplaces. But sometimes OCR simply fails to detect any part before a whitespace character. I feel like it happens more often when it is smaller than usual. For example 'Texas the Omertosa' can be just 'Omertosa.' Since the original codenames of the other alternate operators come first in YoStarEN/YoStarKR, this also can be handled from ocrReplaces. But Yato and Noir Corne are exceptions here, so if the operators cannot be distinguished with the OCR model file that is for YoStarKR currently, we might even have to replace it. Well, of course it is possible that it works fine. We'd need to run some tests and check if it can be done by this.

Constrat:
Damn, you reminded me that I have to check if Yato Alter and Noir Alter work on the english version

178619:
Well, at the very least '야토' and '느와르코르네' should be '^야토$' and '^느와르코르네$', so they won't mess up alter operators. (This is what I have been doing for other operators with alternate forms)

Also, whitespaces are almost never detected with the current OCR model for YoStarKR so I've made it always remove whitespace characters. Therefore Maa would fail to detect Yato/Noir Corne Alter in any case for YoStarKR with the current tasks.json. '키린R 야토' shall be '키린R야토' instead. 

It would be better if we make auto resource update script to remove them for YoStarKR. We can also try to completely change the OCR model for YoStarKR, but it would be a risky plan that will also take a long time and an effort. I also think the results will be rather unstable.

Constrat:
> Well, at the very least '야토' and '느와르코르네' should be '^야토$' and '^느와르코르네$', so they won't mess up alter operators. (This is what I have been doing for other operators with alternate forms)

Yeah, I checked the English and I was missing 5 non alter operators that has alters... Hypergryph is really printing alter after alter. It's almost a joke at this point

Constrat:
I've reverted the probable regression tasks.json commit just to leave the `docs/operators.md` change

PR #6512, opened 2023-09-24
perf: optimize time diff format in sanity report

KevinT3Hu:
The time diff in sanity report in the current version is in the format h:mm which can cause confusion. I made changes to specificly note the units. For example, in Chinese, it would show as x时xx分.

### DIFF
--- a/src/MaaWpfGui/Main/AsstProxy.cs
+++ b/src/MaaWpfGui/Main/AsstProxy.cs
@@ -646,7 +646,7 @@ namespace MaaWpfGui.Main
                             .Replace("{Preset}", configurationPreset);
                         if (isSanityForecastSucc)
                         {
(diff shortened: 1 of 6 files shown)


src/MaaWpfGui/Main/AsstProxy.cs
```diff
@@ -646,7 +646,7 @@ private void ProcTaskChainMsg(AsstMsg msg, JObject details)
                             .Replace("{Preset}", configurationPreset);
                         if (isSanityForecastSucc)
                         {
-                            sanityReport = sanityReport.Replace("{DateTime}", reportTime.ToString("T")).Replace("{TimeDiff}", timeDiff.ToString(@"hh\:mm"));
+                            sanityReport = sanityReport.Replace("{DateTime}", reportTime.ToString("T")).Replace("{TimeDiffH}", timeDiff.Hours.ToString()).Replace("{TimeDiffM}", timeDiff.Minutes.ToString());
```
MistEO, line 649: Using string format instead of replace may be a better choice.

PR #6516, opened 2023-09-24
fix: 尝试修复肉鸽卡在招募界面

Lancarus:


Lancarus:
try to fix #6515

Constrat:
Could this be related to lag/latencies during the recruit moment? Because the stars move to the right slowly, but on the case of a lagspike maa might take the screen too late

Constrat (approved):

Lancarus:
> Could this be related to lag/latencies during the recruit moment? Because the stars move to the right slowly, but on the case of a lagspike maa might take the screen too late
The log in https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues/6515 shows that maa can't recognize rencent template of the recruit starts. So I try to lower its `templThreshold` and cut part of the picture.
The task will try 20 times to find the star, and then reset the Rogue Task Chain. In the log it costs over 20 seconds. I think it doesn't matter for maa catching the template of the screen.
Hope it will works.

PR #6592, opened 2023-09-30
fix: 修复日服傀影肉鸽大盗当头识别错误

MejiroSilence:
天翔ける怪盗・劇団篇
有时会识别成: 天翔ける怪盗・劇団窟， 天翔ける怪盗・劇団篤
image


Constrat (dismissed):
```
            [
                "天翔ける怪盗・劇団.*",
                "天翔ける怪盗・劇団篇"
            ],
            [
                "天翔ける怪盗・劇団篇",
                "大盗当头"
            ],
```
2679 to 2686
This is more scalable and gives more info as to what was modified.

MejiroSilence:
> ```
>             [
>                 "天翔ける怪盗・劇団.*",
>                 "天翔ける怪盗・劇団篇"
>             ],
>             [
>                 "天翔ける怪盗・劇団篇",
>                 "大盗当头"
>             ],
> ```
> 
> 2679 to 2686 This is more scalable and gives more info as to what was modified.

but the other ocr replace rules seem to put the regular expression behind.
like 2555 to 2562
```
            [
                "猛獣小屋",
                "驯兽小屋"
            ],
            [
                "猛.*小屋",
                "驯兽小屋"
            ],
```

and 2855 to 2861
```
            [
                "声なき叫び",
                "无声呼号"
            ],
            [
                "声なき.*び",
                "无声呼号"
            ],
```

Constrat:
Looking at the commits, those were kind of older. It's either:
only the regex or both regex and what is correct because iirc checking for `猛獣小屋` before `猛.*小屋` pretty much doesn't make any difference, so `猛獣小屋` can just directly be removed, but this would mean that we lose what is expected.

Constrat:
Looking at other `ocrReplace `tasks I'm noticing they are like you said, but working on them seems a bit tedious especially if you don't have the game open or the docs with you.
After reviewing all of this, you can just choose. Keep the code consistency but with what can be considered "confusion".

Maybe someone will have the time to change all of them

PR #6599, opened 2023-09-30
style: VS22 Solution Explorer name change

Constrat:
I understand that most of the MAA developers use Chinese as main language but I feel like this little change will not impact the developers, as the new nomenclature is basic programming knowledge (needed for VS22) but will definitely help all of the developers that don't understand Chinese.
This change will not impact the VSCode default environment where no programming knowledge is needed.

image

image


ABA2396:
@MistEO

zzyyyl (approved):

MistEO:
My system language is Chinese, so the filter was automatically created in Chinese. I didn’t notice this detail. Thank you!

PR #6625, opened 2023-10-01
perf: WpfGui自动战斗解析作业后，提示文本移动至最顶端

status102:
- https://github.com/orgs/MaaAssistantArknights/discussions/6600

image

现在新版加完战斗列表之后太长了，完全遮挡了作业信息部分。
或许挪到作业信息之前会好一些？

### DIFF
--- a/src/MaaWpfGui/ViewModels/UI/CopilotViewModel.cs
+++ b/src/MaaWpfGui/ViewModels/UI/CopilotViewModel.cs
@@ -255,8 +255,9 @@ namespace MaaWpfGui.ViewModels.UI
                     return;
                 }
 
-                var doc = (JObject)json["doc"];
+                AddLog(LocalizationHelper.GetString("CopilotTip"));
 
+                var doc = (JObject)json["doc"];
                 string title = string.Empty;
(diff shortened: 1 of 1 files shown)


Constrat:
We can start considering having a completely different window showing the tips, instead of having them in the copilot log.

status102:
> Is it possible to change where the default scroll should be? For example: If there is no copilot file it would show "Tips:" and all the other items from the top. Not like this: image
> 
> If the copilot file is loaded it would show the first line of the copilot info like this: image So that you can scroll down

try and failed :( 
so... help wanted

Constrat:
C# and GUIs, my beloved 💢

Constrat:
> C# and GUIs, my beloved 💢

I've been tinkering a bit, but honestly I can't wrap my head around it.

zzyyyl (approved):

Constrat:
@status102 For now this can work as it already helps a bit. Maybe a day when we get a designer in the team we can tinker a bit more.

status102:
> @status102 For now this can work as it already helps a bit. Maybe a day when we get a designer in the team we can tinker a bit more.

let's put it aside for now. haha

MistEO (approved):

PR #6665, opened 2023-10-02
style: hidden "Manual switch" button

Constrat:
YostarEN does not allow multi-accounting (I'm pretty sure it's the same for JP and KR as they are all under Yostar) the "Manual Switch" button is not really needed.
With this PR I'm completely hiding the Grid.


status102:
I forgot to hide it :(

status102 (approved):

PR #6697, opened 2023-10-04
feat: OCR manipulation params for QuickFormation

### LINKED ISSUE #6680: Operator Formation dirty OCR
### Before raising the issue...

- [X] I understand that Issues are for feedback and problem solving, not trolling the comments section, and will provide as much information as possible to help solve the problem.
- [X] I filled in a short, clear title so that developers could quickly identify the general problem when going through the issue list. Instead of "Some suggestions", "Stuck", etc.
- [X] I have checked the latest beta update and there is no mention of the bug being fixed.

### Description

YostarEN is back with OCR issues (should be lighter)
Kroos can't be recognized because of the "RHODES ISLAND" logo in the background.
image
```
asst::WordOcr [{ ISLAN KroOs: [ 0, 0, 113, 21 ], score: 0.688967 }] by OCR Rec , cost 207 ms
Proced [{ ISLAN KroOs: [ 538, 320, 113, 21 ], score: 0.688967 }]
```
I was wondering if there is a way to implement `maskRange` to help the OCR read only WHITE text:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/2a2c137d827e0f03c8645b397fe012e05e2216f0/src/MaaCore/Vision/Battle/BattleFormationAnalyzer.cpp#L23-L32

@MistEO this brings back the good old times

I've started noticing this issue after the nightly of 4.24.1 (but it may be that I never chose Kroos before)
Everything that happened after this has been a complete nightmare.

Constrat:
I still haven't fully tested on the CN, but it has the same specialParams as `OperBoxNameOCR` so it shouldn't create problems.
Closes #6680

### DIFF
--- a/resource/tasks.json
+++ b/resource/tasks.json
@@ -7354,11 +7354,11 @@
             20
         ],
         "postDelay": 200,
-        "specialParams_Doc": "Used in BattleFormationTask.cpp for OCR manipulation",
+        "specialParams_Doc": "USED FOR YOSTAREN. Used in BattleFormationTask.cpp for OCR manipulation. For CN: OCRerConfig.h default values are used.",
         "specialParams": [
-            160,
-            4,
-            4,
+            140,
+            2,
+            0,
             0
         ]
     },
--- a/resource/tasks.json
+++ b/resource/tasks.json
@@ -7353,7 +7353,14 @@
             107,
             20
         ],
-        "postDelay": 200
+        "postDelay": 200,
+        "specialParams_Doc": "Used in BattleFormationTask.cpp for OCR manipulation",
+        "specialParams": [
+            160,
+            4,
+            4,
+            0
+        ]
     },
     "BattleQuickFormationSkill1": {
         "algorithm": "JustReturn",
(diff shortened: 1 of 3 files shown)


MistEO:
The formation screen is not the same size as the operator box screen. It may be better to set CN to the default value (the original default value of the TemplDetOCRer member variable), at least it will not cause regression problems.

Constrat:
Once I'm home I'll change resource/tasks.json

resource/tasks.json
```diff
@@ -7353,7 +7353,14 @@
...
         ],
-        "postDelay": 200
+        "postDelay": 200,
+        "specialParams_Doc": "USED FOR YOSTAREN. Used in BattleFormationTask.cpp for OCR manipulation. For CN: OCRerConfig.h default values are used.",
+        "specialParams": [
+            140,
+            2,
+            0,
+            0
+        ]
```
Constrat, line 7363: https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/9dd13151708915a1f002329e30144684976b5d8d/src/MaaCore/Vision/Config/OCRerConfig.h#L22-L26

```
"specialParams": [
    bin_threshold_lower,
    bin_expansion,
    bin_left_trim_threshold,
    bin_right_trim_threshold
```

Constrat:
Ready @MistEO unless you have any more suggestions

Issue #6664, opened 2023-10-02
Operator names localizations for overseas


Constrat:
### Before raising the issue...

- [X] I filled in a short, clear title so that developers could quickly identify the general problem when going through the issue list. Instead of "Some suggestions", "Stuck", etc.
- [X] I'm probably sure this is a new feature/suggestion and not a bug encountered (please attach logs if you're not sure)

### The problems you have encountered?

Currently:
- Custom infrastructure
- Copilot 
- OperBox identification

Log print the name of the operator/s only in the zh_CN language (how all the operators are managed in MAA).

Is it feasible to create a method that compares the zh_CN name to the selected client language.

### Any good ideas?

[MaaAssistantArknights\docs\glossary\operators.md](https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/915aa0f53b2bad789f3ae879cb0311aa5b0d12b7/docs/glossary/operators.md) already has the list of operators in a language ordered way.

https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/2a2c137d827e0f03c8645b397fe012e05e2216f0/src/MaaWpfGui/ViewModels/UI/RecognizerViewModel.cs#L553-L560
I'm pretty sure I should be looking at this first but my C++ skills don't really allow me anything this complex, help is definitely welcomed

PR #6726, opened 2023-10-06
feat: operbox name language based localization

Constrat:
My idea in:
- #6664

After a bit of trail and error I came up with this bunch of garbage.
I've tried running the code in language: en_US, on the YostarEN client and it's awful.
<details><summary>This is the result: </summary>
<p>

image
image
</p>
</details> 

This code is completely unoptimized because I'm honestly lacking the knowledge of methods to optimize this.
Also I'm limited only to arrays (which I'm guessing are inefficient as they can be).
I'm hoping for some (a lot) of help


ABA2396:
The returned json contains the id and name, so all you need to do is create a dictionary and match the name of the corresponding language with the id.

Constrat:
Yeah this is my next step. I'm trying to create the dictionary, but I'm trying to be lazy and trying to create it "automated" via csv, excel, and vscode but I'm losing my patience xD

MistEO:
The OperBox identification data comes from https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/resource/battle_data.json and we don't have overseas version.
Just let [the resource updater](https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/tools/ResourceUpdater/main.cpp#L145) support it like [recruitment](https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/tools/ResourceUpdater/main.cpp#L186), we don't need to do anything else.

Constrat:
So same thing for the custom infrastructure "preferred opers:"?

MistEO:
However, the size of this json is not small. We can also modify the resource updater and data reader so that the files in global only have id and name, which saves OTA traffic.

MistEO:
> So same thing for the custom infrastructure "preferred opers:"?

No, for custom infrastructure, it only depends on the json input by the user.

Constrat:
I've tried using the english names in the infrastructure as it's said in the docs but it didn't work.
```
"operators": [
                            "夕",   // 使用 OCR 进行，需要传入对应客户端语言的干员名
                            "令",
                            "凯尔希",
                            "阿米娅",
                            "玛恩纳"
```

MistEO:
> I've tried using the english names in the infrastructure as it's said in the docs but it didn't work.
> To make it work I have to use the simplified ones. (but this is another issue, I'll have a deeper look at that at a later time)

Forget it. The translation of operator names is a pure UI behavior. MaaCore will not know which language is used in the UI, so this translation work should be handled by C#.

MistEO:
> The OperBox identification data comes from [`dev`/resource/battle_data.json](https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/resource/battle_data.json?rgh-link-date=2023-10-06T16%3A25%3A46Z) and we don't have overseas version. Just let [the resource updater](https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/tools/ResourceUpdater/main.cpp?rgh-link-date=2023-10-06T16%3A25%3A46Z#L145) support it like [recruitment](https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/tools/ResourceUpdater/main.cpp?rgh-link-date=2023-10-06T16%3A25%3A46Z#L186), we don't need to do anything else.

Forget it. The translation of operator names is a pure UI behavior. MaaCore will not know which language is used in the UI, so this translation work should be handled by C#.

Constrat:
> Forget it. The translation of operator names is a pure UI behavior. MaaCore will not know which language is used in the UI, so this translation work should be handled by C#.

But how does it work for the recruitment? Shouldn't it just be copy paste?

MistEO:
But hardcoding operator names is still not a good idea. We can use the updater to automatically generate a translation file for UI, which can be JSON or xaml.

ABA2396:
[docs/glossary/operators.md](https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/7af4c6e44b42cde93de44f12478728f1ad6aa0b6/docs/glossary/operators.md?plain=1#L103) has a counterpart, so I'll just take that? But it's a bit strange, actually the "CharsNameOcrReplace" in `tasks.json` has already converted to Chinese, what we need to do now is ocr -> corresponding Chinese name -> id -> corresponding foreign name.

MistEO:
Sorry I just made a mistake and there was some conflict in my answer.

In short, For OperBox, it is correct to solve it by modifying the UI, but we should not hard-code the operator name. We can automatically generate some files needed for translation by the resource updater (mapping of id and multilingual name)

For custom infrastructure, because we use OcrReplace to replace the recognition results with Chinese, Chinese must also be used in the input json. Maybe we should use a separate task's OcrReplace for custom infrastructure, which is English-to-English (instead of English-to-Chinese)

Constrat:
ops.
 Why is it running actions again 😭 


> For custom infrastructure, because we use OcrReplace to replace the recognition results with Chinese, Chinese must also be used in the input json. Maybe we should use a separate task's OcrReplace for custom infrastructure, which is English-to-English (instead of English-to-Chinese)

This looks do-able but for now I would like to concentrate more on the OperBox as I think it would be more useful.
Having the UI tell you what opers are in the base is a bit less important considering the custom infrastructure has to be made first in chinese, so you would know what opers are being selected.

Constrat:
> Operators.md "Jsonned"

I'm guessing this won't be needed if using the autoupdater.

MistEO:
We gave the name in the callback, which may be used by some non-WPF UIs, and the callback depends on battle_data.json. We should try not to delete existing fields. So giving multiple name_fields would be better.

Constrat:
And based on the 

> We gave the name in the callback, which may be used by some non-WPF UIs, and the callback depends on battle_data.json. We should try not to delete existing fields. So giving multiple name_fields would be better.

battle_data.json:
```
"char_002_amiya": {
    "name": "阿米娅",
    "name_EN": "Amiya", 
    "name_JP": "xxxx",
    "name_KR": "xxxx",
    "name_TW": "xxxx",
    "position": "RANGED",
    "profession": "CASTER",
    "rangeId": [
        "3-6",
        "3-1",
        "3-1"
    ],
    "rarity": 5
},
```
RecognizerViewModel.cs:
```
Tuple<string, int> tuple;

switch (ConfigurationHelper.GetValue(ConfigurationKeys.Localization, string.Empty))
{
    case "zh-CN":
        tuple = new Tuple<string, int>((string)operBox["name"], (int)operBox["rarity"]);
        break;

    case "en-US":
        tuple = new Tuple<string, int>((string)operBox["name_EN"], (int)operBox["rarity"]);
        break;

    case "ja-JP":
        tuple = new Tuple<string, int>((string)operBox["name_JP"], (int)operBox["rarity"]);
        break;

    case "ko-KR":
        tuple = new Tuple<string, int>((string)operBox["name_KR"], (int)operBox["rarity"]);
        break;

    case "zh-TW":
        tuple = new Tuple<string, int>((string)operBox["name_TW"], (int)operBox["rarity"]);
        break;
}
```

Constrat:
I'm working on the resourceupdater:
```cpp
    for (auto& [id, char_data] : chars_cn_json.as_object()) {
        json::value char_new_data;
        std::string name = char_data["name"].as_string();
        std::string name_EN = chars_en_json.as_object().char_data["name"].as_string();
        std::string name_JP = chars_jp_json.as_object().char_data["name"].as_string();
        std::string name_KR = chars_kr_json.as_object().char_data["name"].as_string();
        std::string name_TW = chars_tw_json.as_object().char_data["name"].as_string();

        char_new_data["name"] = name;
        char_new_data["name_EN"] = name_EN;
        char_new_data["name_JP"] = name_JP;
        char_new_data["name_KR"] = name_KR;
        char_new_data["name_TW"] = name_TW;
```
Something like this. 
image

Constrat:
Since there are less operators in the global, to manage the non existing id do you guys have any suggestions? Throwing away the exception?

Constrat:
[ba96a79](https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/6726/commits/ba96a7926e1d1477779afe08f3f09d23c966c25d) if you have suggestions on how to fix this. 
in the debug window: 
```
MAA has encountered an error
Object reference not set to an instance of an object.
Solution
Unknown error occurred
Details
System.NullReferenceException: Object reference not set to an instance of an object.
   at MaaWpfGui.ViewModels.UI.RecognizerViewModel.PadRightEx(String str, Int32 totalByteCount) in C:\Users\David\Downloads\MaaAssistantArknights\src\MaaWpfGui\ViewModels\UI\RecognizerViewModel.cs:line 47
   at MaaWpfGui.ViewModels.UI.RecognizerViewModel.OperBoxParse(JObject details) in C:\Users\David\Downloads\MaaAssistantArknights\src\MaaWpfGui\ViewModels\UI\RecognizerViewModel.cs:line 605
   at MaaWpfGui.Main.AsstProxy.ProcSubTaskExtraInfo(JObject details) in C:\Users\David\Downloads\MaaAssistantArknights\src\MaaWpfGui\Main\AsstProxy.cs:line 910
   at MaaWpfGui.Main.AsstProxy.ProcSubTaskMsg(AsstMsg msg, JObject details) in C:\Users\David\Downloads\MaaAssistantArknights\src\MaaWpfGui\Main\AsstProxy.cs:line 704
   at MaaWpfGui.Main.AsstProxy.ProcMsg(AsstMsg msg, JObject details) in C:\Users\David\Downloads\MaaAssistantArknights\src\MaaWpfGui\Main\AsstProxy.cs:line 414
   at MaaWpfGui.Main.AsstProxy.<>c__DisplayClass34_0.<CallbackFunction>b__0() in C:\Users\David\Downloads\MaaAssistantArknights\src\MaaWpfGui\Main\AsstProxy.cs:line 375
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
```
This is because `operBox[localized_name]` is `null` when the GUI.localization is NOT zh-cn

tools/ResourceUpdater/main.cpp
```diff
@@ -795,11 +813,29 @@ bool update_battle_chars_info(const std::filesystem::path& input_dir, const std:
...
     std::map<std::string, std::vector<std::string>> tokens;
-    for (auto& [id, char_data] : chars_json.as_object()) {
+    for (auto& [id, char_data] : chars_cn_json.as_object()) {
         json::value char_new_data;
         std::string name = char_data["name"].as_string();
-
         char_new_data["name"] = name;
+
+        if (chars_en_json.find(id)) {
+            std::string name_en = chars_en_json.as_object()[id]["name"].as_string();
```
MistEO, line 822:
```c++
        if (auto char_en = chars_en_json.find(id)) {
            char_new_data["name_en"] = char_en["name"];
        }
```

MistEO: If not found, do we need to set `char_new_data["name_en"]` to null or id or something?

src/MaaWpfGui/ViewModels/UI/RecognizerViewModel.cs
```diff
@@ -550,9 +550,33 @@ public bool OperBoxParse(JObject details)
...
+
+                default:
+                    localized_name = "name_en";
+                    break;
+            }
+
             foreach (JObject operBox in operBoxes.Cast<JObject>())
             {
-                var tuple = new Tuple<string, int>((string)operBox["name"], (int)operBox["rarity"]);
+                var tuple = new Tuple<string, int>((string)operBox[localized_name], (int)operBox["rarity"]);
```
MistEO, line 579: If a CN client (Arknights) player use MAA in English, when the new operator he has does not have an English name, maybe we can directly display the Chinese name of the operator.

tools/ResourceUpdater/main.cpp
Constrat: If there is no corresponding "global" Id we could just use the CN name. But in EN case, we have the official translations for the operators, the id should also remain the same the problem becomes populating the jsoj for this case.
Or... (This requires more work) check what client is selected and with that we know how many operators there are in that server.
YostarEN has less ops than Official CN

src/MaaWpfGui/ViewModels/UI/RecognizerViewModel.cs
Constrat: Discussion in https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/6726#discussion_r1349474985

tools/ResourceUpdater/main.cpp
Constrat: > ```c++
>         if (auto char_en_opt = chars_en_json.find(id)) {
>             char_new_data["name_en"] = char_en_opt->at("name");
>         }
> ```

Is std::string name needed for the cn? Or we can just remove that variable like you are suggesting for the overseas?

Constrat: > If not found, do we need to set `char_new_data["name_en"]` to null or id or something?

I'm having a look at this at this moment. We could go two ways:
- ResourceUpdater can give the overseas names the CN name if the ID is missing (this way in the UI we should not have anything to worry about)
- or we can have an empty string (ex: `"name_en" = ""`,) and then check in the UI if the names are empty to use the CN name. At the end of the day, the result is the exact same. It will show the CN characater.

Technically, EN has all the names CN has. The problem becomes comparing the two resource jsons, because the EN doesn't have the IDs so it will always give a false.

Constrat:
```json
"char_1016_agoat2": {
    "name": "纯烬艾雅法拉",
    "name_en": "纯烬艾雅法拉",
    "name_jp": "纯烬艾雅法拉",
    "name_kr": "纯烬艾雅法拉",
    "name_tw": "纯烬艾雅法拉",
    "position": "RANGED",
    "profession": "MEDIC",
    "rangeId": [
        "3-3",
        "3-17",
        "3-17"
    ],
    "rarity": 6
},
```
The result is this one.

Constrat:
[0a6ace6](https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/6726/commits/0a6ace6b96c0195042eb5cba74ccfd8c8aa96113) also has testing in it.

Constrat:
It still gives reference to an empty object when using something other than zh_Chinese

Constrat:
`JArray operBoxes = (JArray)details["all_opers"]` does not have the added names:

```json
{[
  {
    "id": "char_002_amiya",
    "name": "阿米娅",
    "own": false,
    "rarity": 5
  },
  {
    "id": "char_237_gravel",
    "name": "砾",
    "own": false,
    "rarity": 4
  },
  {
    "id": "char_107_liskam",
    "name": "雷蛇",
    "own": false,
    "rarity": 5
  },
```

Constrat:
You've gotta be kidding me. The json goes through 5 layers. `OperBoxRecognitionTask.cpp` needs to be edited, and I think I'll need getters

Constrat:
Finally everything now works as expected. the only "issue" I'm noticing is that it's SUPER HEAVY 
image
It's almost as bad as the first two commits, that iterated over and over for every name.

src/MaaCore/Config/Miscellaneous/BattleDataConfig.h
```diff
@@ -24,6 +24,42 @@ namespace asst
             return iter->second.id;
         }
 
+        const std::string get_en(const std::string& name) const
+        { 
+            auto iter = m_chars.find(name);
+            if (iter == m_chars.cend()) {
+                return "";
```
Constrat, line 31: Open to suggestions on what's supposed to go here...

src/MaaCore/Config/Miscellaneous/BattleDataConfig.h
```diff
@@ -24,6 +24,42 @@ namespace asst
...
+            }
+            return iter->second.name_en;
+        }
+
+        const std::string get_jp(const std::string& name) const
+        {
+            auto iter = m_chars.find(name);
+            if (iter == m_chars.cend()) {
+                return "";
```
Constrat, line 40: And here

src/MaaCore/Config/Miscellaneous/BattleDataConfig.h
```diff
@@ -24,6 +24,42 @@ namespace asst
...
+            }
+            return iter->second.name_jp;
+        }
+
+        const std::string get_kr(const std::string& name) const
+        {
+            auto iter = m_chars.find(name);
+            if (iter == m_chars.cend()) {
+                return "";
```
Constrat, line 49: And here

src/MaaCore/Config/Miscellaneous/BattleDataConfig.h
```diff
@@ -24,6 +24,42 @@ namespace asst
...
+            }
+            return iter->second.name_kr;
+        }
+
+        const std::string get_tw(const std::string& name) const
+        {
+            auto iter = m_chars.find(name);
+            if (iter == m_chars.cend()) {
+                return "";
```
Constrat, line 58: And here

Constrat:
Except possible optimizations / better methods (there is definitely possibility, I'm crap at C++) the only that is left is searching for a decent enough Padding 
image

src/MaaWpfGui/ViewModels/UI/RecognizerViewModel.cs
```diff
@@ -623,8 +647,8 @@ public bool OperBoxParse(JObject details)
         /// <summary>
         /// 开始识别干员
         /// </summary>
-        // xaml 中用到了
-        // ReSharper disable once UnusedMember.Global
+        /// xaml 中用到了
+        /// ReSharper disable once UnusedMember.Global
```
ABA2396, line 651: This is not a documentation note, use two '/'

src/MaaWpfGui/ViewModels/UI/RecognizerViewModel.cs
```diff
@@ -550,9 +550,33 @@ public bool OperBoxParse(JObject details)
             List<Tuple<string, int>> operHave = new List<Tuple<string, int>>();
             List<Tuple<string, int>> operNotHave = new List<Tuple<string, int>>();
 
+            string localized_name;
```
ABA2396, line 553: 在 C# 中用的是驼峰命名，这里应该是 "localizedName"

src/MaaWpfGui/ViewModels/UI/RecognizerViewModel.cs
Constrat: The warning pissed me off, my bad

Constrat:
Wrong branch
...

tools/ResourceUpdater/main.cpp
```diff
@@ -763,17 +773,32 @@ bool generate_english_roguelike_stage_name_replacement(const std::filesystem::pa
...
+    // auto& chars_kr_json = chars_kr_opt.value();
+    // auto& chars_tw_json = chars_tw_opt.value();
+
+    std::vector<std::pair<json::value, std::string>> chars_json = { { chars_cn_opt.value(), "name" },
+                                                                    { chars_en_opt.value(), "name_en" },
+                                                                    { chars_jp_opt.value(), "name_jp" },
+                                                                    { chars_kr_opt.value(), "name_kr" },
+                                                                    { chars_tw_opt.value(), "name_tw" } };
+
     auto& range_json = range_opt.value();
```
Constrat, line 802: This is probably going overkill. This is scalable, but I don't think new clients are ever gonna get released
```cpp
    if (!range_opt) {
        return false;
    }

    std::vector global = { "en", "jp", "kr", "tw" };
    std::vector chars_opt = { json::open(input_dir / "gamedata" / "excel" / "character_table.json") };

    for (int i = 1; i < 5; i++) {
        chars_opt.push_back(json::open(overseas_dir / global[i - 1] / "gamedata" / "excel" / "character_table.json"));
    }

    for (auto& server : chars_opt) {
        if (!server) {
            return false;
        }
    }

    std::vector<std::pair<json::value, std::string>> chars_json;
    std::vector name_vector = { "name", "name_en", "name_jp", "name_kr", "name_tw" };

    for (int i = 0; i < chars_opt.size(); i++) {
        chars_json.push_back({ chars_opt[i].value(), name_vector[i] });
    }
```

src/MaaCore/Config/Miscellaneous/BattleDataConfig.h
MistEO: Return the name directly?

Constrat: I've just copied how the getter was for the ID... Didn't really give much thoughts about it

tools/ResourceUpdater/main.cpp
MistEO: > > ```c++
> >         if (auto char_en_opt = chars_en_json.find(id)) {
> >             char_new_data["name_en"] = char_en_opt->at("name");
> >         }
> > ```
> 
> Is std::string name needed for the cn? Or we can just remove that variable like you are suggesting for the overseas?

OK, we can remove `as_string`

Constrat: Yep, did that yesterday.

MistEO: An empty string is probably not a good idea, it means there is a value here (albeit an empty value), not that there is no value. I think

- The callback given by core does not contain name_en. If the UI checks that there is no name_en field, the name field is used.
- The callback given by core always contains name_en, but it may be Chinese, and the UI uses it directly.

Either way is possible, it’s up to you.

Constrat: Is there the "best performance solution"? Because I've noticed that it's a pretty demanding task

MistEO (approved):

Constrat:
Probably not the easiest start for C++ programming xD.
How would you rate this PR

MistEO:
No problem XD

Constrat:
I'll change the getter. Give a last check and then merge

Constrat:
Crap

MistEO:
> Finally everything now works as expected. the only "issue" I'm noticing is that it's SUPER HEAVY image It's almost as bad as the first two commits, that iterated over and over for every name.

because of this: https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues/5095

Constrat:
But this was during operbox recognition, not resource updater

MistEO:
Oh, I misunderstood xD

src/MaaCore/Config/Miscellaneous/BattleDataConfig.cpp
```diff
@@ -12,7 +12,16 @@ bool asst::BattleDataConfig::parse(const json::value& json)
...
+        std::string name_en = char_data_json.at("name_en").as_string();
+        std::string name_jp = char_data_json.at("name_jp").as_string();
+        std::string name_kr = char_data_json.at("name_kr").as_string();
+        std::string name_tw = char_data_json.at("name_tw").as_string();
+
         data.name = name;
+        data.name_en = name_en;
+        data.name_jp = name_jp;
+        data.name_kr = name_kr;
+        data.name_tw = name_tw;
```
Constrat, line 24: @MistEO 
```cpp
        data.name = name;

        data.name_en = char_data_json.at("name_en").as_string();
        data.name_jp = char_data_json.at("name_jp").as_string();
        data.name_kr = char_data_json.at("name_kr").as_string();
        data.name_tw = char_data_json.at("name_tw").as_string();
```

src/MaaCore/Config/Miscellaneous/BattleDataConfig.h
```diff
@@ -24,6 +24,42 @@ namespace asst
...
+            }
+            return iter->second.name_tw;
+        }
+
```
Constrat, line 62:
```cpp
        const std::string get_en(const std::string& name) const { return m_chars.find(name)->second.name_en; }

        const std::string get_jp(const std::string& name) const { return m_chars.find(name)->second.name_jp; }

        const std::string get_kr(const std::string& name) const { return m_chars.find(name)->second.name_kr; }

        const std::string get_tw(const std::string& name) const { return m_chars.find(name)->second.name_tw; }
```

src/MaaCore/Config/Miscellaneous/BattleDataConfig.cpp
Constrat: Nevermind, looking at the elapsed time, this is much slower

PR #6749, opened 2023-10-08
fix: update UnlockClues for EN

Enochen:
The OCR currently fails for the `UnlockClues` task at higher resolutions, so I lowered the threshold a bit for EN (from default `0.90` to `0.85`).

Logs for `1600x900` client:

```
match_templ | UnlockClues.png score: 0.871378 rect: [ 554, 619, 256, 62 ] roi: [ 200, 600, 700, 120 ]
```

### DIFF
--- a/resource/global/YoStarEN/resource/tasks.json
+++ b/resource/global/YoStarEN/resource/tasks.json
@@ -2497,6 +2497,9 @@
             "Unable to receive new Clues"
         ]
     },
+    "UnlockClues": {
+        "templThreshold": 0.85
+    },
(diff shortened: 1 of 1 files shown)


Constrat:
I've noticed in both PR that you are using 1600x900. Unfortunately I've not been playing in that resolution. I'm starting to think that the issues are arising because of that. 
Lowering the threshold should not make any breaking changes especially because it's just a 0.05 change

Constrat (approved):

MistEO:
Thx-ld

PR #6750, opened 2023-10-08
fix: update InfrastReward task (Max Trust base notification) for EN

Enochen:
Currently the `InfrastReward` task in EN looks for the string `"Trust"`. However, the OCR actually parses the text on screen as `"Max Trust"` so the task is unable to progress.

Logs for `1600x900` client:

```
asst::WordOcr [{ Backlog: [ 49, 83, 68, 26 ], score: 0.995247 }, { Max Trust: [ 218, 79, 55, 15 ], score: 0.964198 }, { MAX: [ 169, 92, 37, 16 ], score: 0.994455 }, { reached: [ 219, 99, 44, 13 ], score: 0.998339 }] by OCR Pipeline , cost 276 ms
```

<details>
<summary>Screenshot containing icon on EN client</summary>

[Source](https://www.reddit.com/r/arknights/comments/wzhuju/theres_a_max_trust_notification_for_base/)
image
</details>

### DIFF
--- a/resource/global/YoStarEN/resource/tasks.json
+++ b/resource/global/YoStarEN/resource/tasks.json
@@ -2420,7 +2420,7 @@
         "text": [
             "Collectable",
             "Orders Acquired",
-            "Trust"
+            "Max Trust"
         ]
     },
     "ShamareThumbnail1": {


MistEO:
`Max Trust` will only appear when the operator trust in the control center is full, it is just a reminder.
`Trust` is the trust of any room operator. It is updated at 4 o'clock every day. Clicking it can increase the trust value of all operators.
These two are not the same button.

ABA2396:
"Trust" is the once-a-day one, and "Max Trust" is the prompt after a Control Center agent has 100 trust. Change it to
```
    "InfrastReward": {
        "fullMatch": true,
        "text": [
            "Collectable",
            "Orders Acquired",
            "Trust",
            "Max Trust"
        ]
    },
```

MistEO:
And also, this is a substring matching. If `Max Trust` is on the screen, we should be able to match it using `Trust`. What's the real problem?

ABA2396:
> And also, this is a substring matching. If `Max Trust` is on the screen, we should be able to match it using `Trust`. What's the real problem?

那个任务加了全字匹配，而其这个就算不点也没关系吧

Constrat:
The change I made is wanted. If we check for `Max Trust` it will get stuck trying the click the notification, but that notification is not clickable

Constrat:
I think I've now understood what you meant in the first comment:
the OCR reads `Max Trust` and that is correct, the OCR reads the entire bottom line.
```json
        "roi": [
            0,
            600,
            800,
            118
        ],
```
`Trust` is the two-times daily trust reward that NEEDS to be clicked, when clicked it will go away.
`Max Trust` is just a notification, clicking it, is useless and will not do anything. To make it go away you need to switch the operator.

I've added `fullMatch` because before, whenever the `Max Trust` notification came up, MAA would try to click, without any effect.
@ABA2396 @Enochen

PR #6794, opened 2023-10-08
perf: not stationed before oper selection

Constrat:
In the dormitories infrastructure task, MAA will enter the first dorm, scan for operators, choose 5 of them and only then it will go in the sorting menu to toggle `"Not Assigned"`.
This way for `Dorm 1`, `2` and `3` it won't have to scan as many operators as in the first dorm.

Why is it this way? To make it a bit more efficient shouldn't we just `"Not Assigned"` first and then scan and choose operator.

This PR is at first a question.
If this was indeed a oversight, this PR is here to optimize. If it wasn't an oversight, I would love for someone to enlighten me as why is it this way.


Constrat:
I finally had the time to fully test this.
It works as expected and I don't see any contraindications as to not put it before.

MistEO:
For custom operators, they may be not `Not Assigned`. That may be the case, but to be honest, I don’t know_(:з」∠)_

Constrat:
But this is not consistent. In the case of an operator conflict because they are already assigned, this would only happen on the dorm 0, not in 1,2,3.
This increase custom base  consistency (if the option is selected obviously)
And as I said before, it would reduce time and performance cost, as the first dorm won't have to check through all operators (considering Arknights shows you the lowest morale first, which are usually stationed operators)

MistEO:
Yes, after turning on the `Not Assigned` option, our custom operators do not support the second and third dormitories...

PR #6901, opened 2023-10-13
feat: Only check version.json for Resource OTA

MistEO:
close #6881


Constrat:
@MistEO to go against the issues created by #7418 in [res-update-game.yml](https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/7418/files#diff-8b8245c054ae992ad291b8daaa68d6524bdd0431c1f1fd8a8b2d6debc52cf165), if I manage to "update" the     `result["last_updated"] = time;` only when a change is made to version.json instead of updating it everytime the ResourceUpdater is ran, will this break something?

ABA2396:
1f703b75a8f7169c77c5f7652d05c3db276f2725 Was there ever a case of updating version without updating anything else?

Constrat:
The correct fix would be to update the `last_updated` exactly only when a change is made, instead of just calling the function `asst::utils::get_format_time()` everytime, this way we can correctly just check for only sorted.
(check only if sorted is also technically wrong, because if I just rewrite a line (instead of removing / adding lines) it will consider it as a sorting and not commit the changes. I made it this way to not commit the everytime the `last_updated` gets changed.

Issue #6966, opened 2023-10-17
Save Debug_Img EN shop trader


Constrat:
### Before raising the issue...

- [X] I filled in a short, clear title so that developers could quickly identify the general problem when going through the issue list. Instead of "Some suggestions", "Stuck", etc.
- [X] I'm probably sure this is a new feature/suggestion and not a bug encountered (please attach logs if you're not sure)

### The problems you have encountered?

@zzyyyl like you did for the battleformation of credit fight, can you add a debug screenshot for the Mizuki I.S. the first time it enters the trader?
There are so many items that still need to get OCR for EN that I can't just look at it while running. Having them as screens would help this a lot.
Maybe this "feature" is only available if it's a debug version/debug.txt as to not flood other users debug/Roguelike folder

### Any good ideas?

Something similar to 
- 3674f0cf774e77c1d30072af8d95e9cdc618b80c
- a3f0466

PR #6973, opened 2023-10-17
feat: 增加任务截图插件

zzyyyl:
link to #6966 
link to #6975 (this PR is not a fix for it, but to add debug screenshots for similar issues)

这个插件可以被用在任意 Task，现在只有 RoguelikeTask 注册了

### DIFF
--- a/resource/global/YoStarEN/resource/tasks.json
+++ b/resource/global/YoStarEN/resource/tasks.json
@@ -3316,6 +3316,11 @@
     "StageDrops-DropType-Reward": {
         "templThreshold": 0.7
     },
+    "ScreenshotTaskPlugin-Config-Debug": {
+        "next": [
+            "Roguelike@TraderRandomShopping"
+        ]
+    },
     "Roguelike@ChooseDifficultyConfirm": {
         "text": [
             "Select"
--- a/resource/tasks.json
+++ b/resource/tasks.json
@@ -8586,6 +8586,18 @@
             20
         ]
     },
+    "ScreenshotTaskPlugin-Config": {
+        "Doc": "本任务是插件 ScreenshotTaskPlugin 的配置",
+        "Doc_2": "需要在任务 Task 开始时截图时，将 Task 加入到此任务的 next 列表。", 
+        "algorithm": "JustReturn",
+        "next": []
+    },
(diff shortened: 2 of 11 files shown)


Constrat:
So it will take a screen of what's in `next`?
Looks very scalable also for other debugging purposes.

Constrat:
So it will take a screen of what's in `next`?
Looks very scalable also for other debugging purposes.

Constrat:
Misclick

Constrat:
Something like this?
Let's just hope `next` works without having to copy the task from the CN.

zzyyyl:
@Constrat
I think it's better not to save screenshots of the Trader Store in the release (for all users), you can add them locally if you need to.

see e4f8f1839ba1f3c15b52abbaf8d99aae28daf99e

Constrat:
The idea as to have them only for debug users

Lancarus:
I agree to Constrat. It may be mainly used by debug users. 
Could it run just in debug mode or beta version？
And It may be used in other function or situation，not only in roguelike mode？

status102:
感觉可以按照检查更新的版本，如果是内测或者公测就启用。

zzyyyl:
现在有两个截图，一个是 release 下的截图，一个是 存在 "DEBUG" 或者 "DEBUG.txt" 时的额外截图

Constrat (approved):

zzyyyl:
重构了一下：）

现在这个插件可以被用在任意 Task，不过现在只有 RoguelikeTask 注册了

Lancarus:
但是为啥要在商店加截图

Lancarus (approved):

Constrat:
> 但是为啥要在商店加截图，商店现在应该没啥bug？

Some store items names have two lines.
The OCR algorithm still has no way to interprent
`texttext\ntexttext`

Lancarus:
> Some store items names have two lines. The OCR algorithm still has no way to interprent `texttext\ntexttext`
> 
> Because of this I have to know how all the items are rendered in the store, to regex them I haven't found a solution yet apart from this...

I see...That's really a problem.Localization is really a troublesome job.
Fortunately, we do have you.

PR #6995, opened 2023-10-18
chore: 重构 tools/AutoLocalization

quyansiyuanwang:
其中包含部分更优化和跟规范的算法或结构


quyansiyuanwang:
发现translate.py中ChatTranslator的translate方法改后有错误(不符原逻辑)！
waiting for fixing
labeled 2023.10.20

Constrat:
I'll have a look at this at home, reviewing code by phone is not ideal, I also want to test it

Constrat:
@quyansiyuanwang  `os.path.exists` won't detect .env in my workspace.
Manually `python`ing the termianl then `os.path.exists('.env')` works.

tools/AutoLocalization/src/auto_localization/cli.py
```diff
@@ -10,11 +10,13 @@
...
 logging.debug(os.path.abspath('.env'))
-load_dotenv(dotenv_path='.env')
-if os.path.exists('../.env'):
-    load_dotenv(dotenv_path='../.env')
-if os.path.exists('../../.env'):
-    load_dotenv(dotenv_path='../../.env')
+
+for t in range(3):
+    env_path = t * '../' + '.env'
+    if os.path.exists(env_path):
```
Constrat, line 16: I've tested some stuff but couldn't really get my head around it.
Does it work on your machine?

Constrat (changes requested):

tools/AutoLocalization/src/auto_localization/cli.py
quyansiyuanwang: but `os.path.exists` is from original code, do it work rightly in original code?

quyansiyuanwang: @UniMars 你的这个`os.path.exist`真的能正常运行吗，你当时测试结果如何

quyansiyuanwang: I've asked gpt3.5 for why
it told me should splicing it by `join` but I think that's not the problem

Constrat: My bad, debugging it and / or running py from a different folder won't make it work.  `os.path.exists` starts from the directory where you ran the file, not the running directory (I wasn't expecting this)

Constrat (commented):
It seems pretty wonky (but this is before your changes):
- to make use `LOCALIZATION_PATH=src\MaaWpfGui\Res\Localizations` you need to run the python file from the repo parent directory (MaaAssistantArknights) aka: `py tools\AutoLocalization\run.py`
- but to use the `.env` file you need to run  the python file from MaaAssistantArknights\tools\Autolocalization so: `py run.py`

To make it work as-is `.env` needs to be placed in the root of the repository which is not really ideal

EDIT: This is a pain especially when working with VSCode
You can right click the .py file and run it without commands, but it will use the parent directory as main directory, and
```py
for t in range(3):
    env_path = t * '../' + '.env'
    if os.path.exists(env_path):
        load_dotenv(dotenv_path=env_path)
        break
```
does not allow it.

Constrat:
This is ugly as it can get, I know but this way if the script is ran from the parent directory it won't crash.

The problem is: the only way to make the script work is if it's ran from the parent directory (MaaAssistantArknights) because the `LOCALIZATION_PATH` is not scalable

Constrat:
To manage the possibility of different parent directories I have no idea.

tools/AutoLocalization/src/auto_localization/cli.py
quyansiyuanwang: Okay, I didn't expect this situation for a moment either.

quyansiyuanwang:
> This is ugly as it can get, I know but this way if the script is ran from the parent directory it won't crash.
> 
> The problem is: the only way to make the script work is if it's ran from the parent directory (MaaAssistantArknights) because the `LOCALIZATION_PATH` is not scalable
> 
> To manage the possibility of different parent directories it would need to be scalable in two ways such as going back through the directories to reach the parent.
> 
> Implementing a parent directory (that would always be that one [MaaAssistantArknights] would be the idea )
> 
> EDIT: I was looking at some other code, this is cpp and the idea is that it would iterate until it finds the parent directory.
> 
> ```c++
>     auto solution_dir = cur_path;
>     for (int i = 0; i != 10; ++i) {
>         solution_dir = solution_dir.parent_path();
>         if (std::filesystem::exists(solution_dir / "resource")) {
>             break;
>         }
>     }
> ```
> 
> @quyansiyuanwang is this possible in Python?


Yes,there is a solution by using pathlib.
In the end, the solution_dir will stoped at the parent directory.
Test it and check its result is truly you want. 
python below:
```python
from pathlib import Path

solution_dir = Path.cwd()
for _ in range(10):
    solution_dir = solution_dir.parent
    resource_dir = solution_dir / "resource"
    if resource_dir.exists():
        break
```

Constrat:
Looks very clean. I'll check everything and if it works as expected I'll merge it.
Do you need any more refactoring? Let me know :)

quyansiyuanwang:
I have tried my best to make the code clean and reasonable.
So I think there is not anymore for me to restruct or imporve.
Even if there were, I wouldn't bother to change it,'cause Iam lazy,hahaha

Constrat (approved):
Approving but the only way to

quyansiyuanwang:
oh, great changes of formating.
I love the style like this.
Nice job,bro.
hahaha

Constrat:
> oh, great changes of formating. I love the style like this. Nice job,bro. hahaha

Shift + Alt + F for the win.
I'm merging.
Thank you for your contribution!

PR #7212, opened 2023-11-05
feat: 水月肉鸽烧水更多奖励

Manicsteiner:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues/7048
支持在水月肉鸽起始选择更多烧水奖励，避免选择减希望。我这里跑了两次，能用。合并前还要替换掉 resource/template 那两张图。


Constrat:
Does the template need to include the text? Considering we are not using WordOcr wouldn't just identifying the key (and the triangle) be enough? So that we don't have to localize images.

Manicsteiner:
> Does the template need to include the text? Considering we are not using WordOcr wouldn't just identifying the key (and the triangle) be enough? So that we don't have to localize images.
It is true that only pictures are enough, but doing in this way requires modifying all ROIs and related pictures. I will try that later.

Constrat:
Technically the ROI should not be a problem (it would just waste performance) but it should work (Issues arise when the ROI is SMALLER than the image, and this is not the case as we are reducing the dimensions of the images). Modifying all the pictures is something that I can do if you give me the list of all the pictures.
I could actually change the ROI if I have all the images I need to modify.

Manicsteiner:
> Technically the ROI should not be a problem (it would just waste performance) but it should work (Issues arise when the ROI is SMALLER than the image, and this is not the case as we are reducing the dimensions of the images). Modifying all the pictures is something that I can do if you give me the list of all the pictures. I could actually change the ROI if I have all the images I need to modify.


You are right. I tried just cropping out the text at the bottom of the LastReward related pictures and deleting the corresponding translated version in resource\global\YoStarJP\resource\template, and it still works fine.
Related pictures are ```Mizuki@Roguelike@LastReward.png, Mizuki@Roguelike@LastReward2.png, Mizuki@Roguelike@LastReward3.png, Mizuki@Roguelike@LastReward4.png, Mizuki@Roguelike@LastReward5.png, Mizuki@Roguelike@LastReward6.png``` in resource\template. Translated versions in resource\global\\*server\*\resource\template . 
```Phantom@Roguelike@LastReward``` and ```Sami@Roguelike@LastReward``` may also need this change.
And here are original screenshots, in case of you need it.
Screenshot_1698074544
Screenshot_1698075992

Constrat:
Perfect you still have the original images !!! I'll have lunch then commit the new images with the corresponding ROI

Constrat:
Do you think something like this could work?
cc @Lancarus @zzyyyl

zzyyyl:
冲突有点多，改完再开一个吧

PR #7253, opened 2023-11-09
fix: 修复在部分情况下，自动战斗-自动编队时干员页面最边缘干员名字未露出，导致编队时表现错误

status102:


Constrat:
I think something similar is needed in OperBox algorithm. Everytime you run it, when it gets to the end it will try to scroll multiple other times (but it won't go anywhere as there no other operators left)

ABA2396:
> I think something similar is needed in OperBox algorithm. Everytime you run it, when it gets to the end it will try to scroll multiple other times (but it won't go anywhere as there no other operators left)

OperBox does close the career bar, this one doesn't. Slide it a few times to make sure it recognizes the last person.

status102:
I don't know if there is a similar problem in OperBox. It seems that there are some issues may be caused by the roi of `BattleQuickFormation-OperNameFlagBase`. When role icon is close to screen edge, ocr result may be empty.

status102:
可能得到空值/错误，但是去测一个不会错误识别的范围有点麻烦，就这样吧

PR #7328, opened 2023-11-16
feat: SSS#2 for YostarEN

Constrat:


Constrat:
The regex order for the ocr replace is:
Mama John:
LT-1, LT-2, LT-3, LT-4, LT-5, LT-6
Raytheon Industries
LT-1, LT-2, LT-3, LT-4, LT-5, LT-6

Manicsteiner:
You can open the battle only once and click on the arrows above and below the stage name to view all six stage names. And "Expanded tactical unit"(I don't know it's name in EN server) also only need to select it at the start and back to preview page, no need to no need to wait until stage-4.

Constrat:
Beautiful, thank you very much.
Iirc the A or B routes for equipment are usually not used

Constrat:
@zzyyyl @ABA2396 Just an opinion.
Should I remove the older SSS stages from SSSStageNameOCR, as they are not available as game content anymore?
I'm worried (because of EN `\n` issue I could get duplicates so I was more inclined in removing them (as I can't comment them out)

Constrat:
ops...

Constrat:
Closed because I fucked up. Replaced by:
- #7346

PR #7473, opened 2023-12-04
feat: wpf设置-连接设置新增近期截图耗时

status102:
每10次截图更新一次，计算近30次的min/avg/max
image

PR #7476, opened 2023-12-04
perf: reduced screencap timeout

Constrat:
refs:
- https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/7473#issuecomment-1838440548
- https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/7473#issuecomment-1838446255

### DIFF
--- a/src/MaaCore/Controller/AdbController.cpp
+++ b/src/MaaCore/Controller/AdbController.cpp
@@ -378,7 +378,7 @@ bool asst::AdbController::screencap(cv::Mat& image_payload, bool allow_reconnect
 
         auto start_time = high_resolution_clock::now();
         if (m_support_socket && m_server_started &&
-            screencap(m_adb.screencap_raw_by_nc, decode_raw, allow_reconnect, true)) {
+            screencap(m_adb.screencap_raw_by_nc, decode_raw, allow_reconnect, true, 5000)) {
             auto duration = duration_cast<milliseconds>(high_resolution_clock::now() - start_time);
             if (duration < min_cost) {
                 m_adb.screencap_method = AdbProperty::ScreencapMethod::RawByNc;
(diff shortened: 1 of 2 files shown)


ABA2396:
AdbController.h:
``` cpp
bool screencap(const std::string& cmd, const DecodeFunc& decode_func, bool allow_reconnect = false,
               bool by_socket = false, int time_out = 20000);
```
AdbController.cpp:
``` cpp
bool asst::AdbController::screencap(const std::string& cmd, const DecodeFunc& decode_func, bool allow_reconnect,
                                    bool by_socket,int time_out)
{
    if ((!m_support_socket || !m_server_started) && by_socket) [[unlikely]] {
        return false;
    }

    auto ret = call_command(cmd, time_out, allow_reconnect, by_socket);
```
``` cpp
screencap(m_adb.screencap_raw_by_nc, decode_raw, allow_reconnect, true, 5000)) {
```

I mean something like this? Maybe check to see if it's affecting other areas

Constrat:
Oh, I directly changed the call funcion.
I'll add it as a parameter 👌🏻

ABA2396:
The call_command is used for more than just screenshots, and I'm not sure if I'll run into any other problems if I change the default value of this directly. I don't think we should just treat it as 5000 after the timeout, after the timeout is over is the time used for screencap_raw_by_nc still 5000ms?

ABA2396 (approved):

status102:
#7189  In some cases screencap cost more than 5 seconds, and even 10 seconds in some extreme cases

ABA2396:
> #7189 In some cases screencap costs more than 5 seconds, and even 10 seconds in some extreme cases.
> 
> Sometimes the cost is double than normal with copilot module running. I don't know why it happened

这种估计用啥都不好使了）就算第一个能用用另外两个也没差，主要是timeout的时候应该返回false而不是RawByNc cost 5000 ms吧？

status102:
我觉得只要不用自动战斗、肉鸽，截一次图一分钟都无所谓（

> 主要是timeout的时候应该返回false而不是RawByNc cost 5000 ms吧？
#7473 晚点改改，#7441 看上去应该不受影响

ABA2396:
``` cpp
    auto ret = call_command(cmd, timeout, allow_reconnect, by_socket);

    if (!ret || ret.value().empty()) [[unlikely]] {
        Log.warn("data is empty!");
        return false;
    }
```
``` cpp
    exit_res = m_platform_io->call_command(cmd, recv_by_socket, pipe_data, sock_data, timeout, start_time);

    if (!exit_res) {
        return std::nullopt;
    }
```
@Constrat  There are two implementations of call_command, but AdbLite doesn't yet support timeout, does turning it on have any effect?

Constrat:
@ABA2396 The call_command already has the timeout value in the parameters field.
```cpp
    protected:
        std::optional<std::string> call_command(const std::string& cmd, int64_t timeout = 20000,
                                                bool allow_reconnect = true, bool recv_by_socket = false);
```
So, if it worked before, it will definitely work now. I haven't touched anything except the numbers.

PR #7485, opened 2023-12-05
perf: removed unused templates

Constrat:
17c3911 generalized it
40f38b8 ulterior fixes
@ABA2396 if this does work on EN it will work on the other clients as well


ABA2396:
If you can, look for an Apple-side user so that the number of stones he has on the Android side is 0. This is the case for the national client, I don't know if it's the same for the foreign client

Constrat:
Negative, all the players I know use Android 
(Using Mac is suicide)

PR #7492, opened 2023-12-06
fix: 尝试修复EN服狂人号任务检查错误

status102:


status102:
@Constrat It seems a problem about resource loading process.

Steps:
1. Load official resource
2. Load YoStarJP
3. YoStarEN
4. YoStarKR
5. txwy

When task type changed in former server, latter server would be broken

Constrat:
I think it has to do with the synchronization.
In debug mode it follows a strict procedure to load them one by one.
It seems like it should FIRST load the official , then all of the other.
Right now it's loading JP first, and then official and then EN etc

status102:
I didn't notice that it loaded JP first. But loading multi global server resource may be a more serious mistake.

PR #7511, opened 2023-12-08
ci: 修复smoke-testing重复加载外服和国服资源导致潜在的非预期TaskData字段问题

status102:
修复类似 [0291163](https://github.com/MaaAssistantArknights/MaaAssistantArknights/commit/02911638ce49730d3745fd910c250bc31a9cd44c) 的问题

（本地DEBUG就先扔着了

### DIFF
--- a/src/Cpp/main.cpp
+++ b/src/Cpp/main.cpp
@@ -20,7 +20,18 @@ int main([[maybe_unused]] int argc, char** argv)
         return -1;
     }
 
-#ifdef ASST_DEBUG
+#ifdef SMOKE_TESTING
+    if (argc >= 2) {
+        std::string overseas_type(argv[1]); // "YoStarJP", "YoStarEN", "YoStarKR", "txwy"
+        std::cout << "load overseas_type:" << overseas_type << std::endl;
+
+        const auto overseas_path = cur_path / "resource" / "global" / overseas_type;
+        if (!AsstLoadResource(overseas_path.string().c_str())) {
+            std::cerr << "-------- load resource failed: " << overseas_type << " --------" << std::endl;
+            return -1;
+        }
+    }
+#elif defined(ASST_DEBUG)
     bool load_error = false;
     const auto overseas_dir = cur_path / "resource" / "global";
     for (const auto& client : { "YoStarJP", "YoStarEN", "YoStarKR", "txwy" }) {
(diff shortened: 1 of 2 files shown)


Constrat:
@status102 Is it normal that Sample.exe in Release is 212KB while in Debug it's 72KB?
Also, the release one shows all the outputs correctly (as expected). The debug only opens a terminal and then closes quickly.

Constrat (approved):
`DEBUG`:
- [x] YostarEN 
- [x] YostarJP
- [x] YostarKR 
- [x] txwy 
I'm guessing there is "Official"

Constrat (changes requested):
I want to try and make some changes.
Wait before merge :)

status102:
wow, but it will look like `Sample.exe` (with no param)

Constrat:
> wow, but it will look like `Sample.exe` (with no param)

Yeah just need to add a check: if there is a param, don't run the official loader

status102:
> Yeah just need to add a check: if there is a param, don't run the official loader

no, it is no sense. You need to load official resource FIRST. Overseas resource is incomplete.

Constrat:
> > Yeah just need to add a check: if there is a param, don't run the official loader
> 
> no, it is no sense. You need to load official resource FIRST. Overseas resource is incomplete.

So 4 official needs to be loaded 4 times? For every smoketest?

status102:
> So `official` needs to be loaded 4 times? For every smoketest?

yes

Constrat (approved):

status102:
> So `official` needs to be loaded 4 times? For every smoketest?

Overseas resource files show resource differences between official and overseas. So we have to loaded more than once.

src/Cpp/main.cpp
```diff
@@ -16,11 +16,22 @@ int main([[maybe_unused]] int argc, char** argv)
     // 这里默认读取的是可执行文件同目录下 resource 文件夹里的资源
     bool loaded = AsstLoadResource(cur_path.string().c_str());
     if (!loaded) {
-        std::cerr << "-------- load resource failed --------" << std::endl;
+        std::cerr << "-------- load resource failed: official --------" << std::endl;
         return -1;
     }
 
-#ifdef ASST_DEBUG
+#ifdef SMOKE_TESTING
```
zzyyyl, line 23: 这玩意在哪定义了吗

status102: 没找到（不知道是怎么算的，但是能跑，很怪

status102: https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/src%2FCpp%2FMaaSample.vcxproj#L208-L220

status102:
没问题就合了 @zzyyyl

PR #7523, opened 2023-12-10
perf: SSS Stage Name standardization

Constrat:
This will probably require some testing before merging.
I've tested YostarEN and it seems to be working correctly, tomorrow I'll test a bit more intensively.
I can try to test in the Official client but I'm not sure I can reach the maps, because the account is not levelled enough.
I don't have a JP, KR and txwy account. 
- [ ] Official
- [x] YostarEN
- [ ] YostarJP
- [ ] YostarKR
- [ ] txwy


Constrat:
This will probably require some testing before merging.
I've tested YostarEN and it seems to be working correctly, tomorrow I'll test a bit more intensively.
I can try to test in the Official client but I'm not sure I can reach the maps, because the account is not levelled enough.
I don't have a JP, KR and txwy account. 
- [ ] Official
- [ ] YostarEN
- [ ] YostarJP
- [ ] YostarKR
- [ ] txwy

status102 (changes requested):
我打算调整一下识别名字的逻辑，稍等一下
- link to #7525

status102:
Another problem, roi is too small

image

Constrat:
Stale.
Switching to #7527 for continuation.

PR #7527, opened 2023-12-11
perf: 优化保全关卡识别逻辑，改为使用LT-N进行识别

status102:
- link to #7523 

这样就不用修改过往的作业了


Constrat:
So no need to change all the Stage names for the copilot/SSS_xxx.json?
I can just close #7523 if you copied / transfered everything that was needed @status102

status102:
> So no need to change all the Stage names for the copilot/SSS_xxx.json?

No, we use map json files to get stage code (LT-1 ~ LT-6) now. So needn't change copilot files unless YJ, who publish Arknight, changed map json files.

Constrat:
> > So no need to change all the Stage names for the copilot/SSS_xxx.json?
> 
> No, we use map json files to get stage code (LT-1 ~ LT-6) now. So needn't change copilot files unless YJ, who publish Arknight, changed map json files.

So we use the Official `Stage_name` to "convert" to `LT-n` through the `stages.json`?

status102:
> What `map.json`?

In `resource\Arknights-Tile-Pos`.

```
{
    "code": "GT-1",
    "height": 7,
    "levelId": "activities/a001/level_a001_01",
    "name": "日正当中",
    "stageId": "a001_01",
    "tiles": [
```
levelId, name, stageId. All of them are valid.

Constrat:
Oooh, `arknights tile pos`, I forgot about that `JSON`.
👍🏻

Constrat:
image
ROI seems to be good for EN (that has "smaller numbers")
If it works for official it should definitely work for JP, KR and txwy.

PR #7764, opened 2023-12-23
chore: YoStarKR missing item modfied

HX3N:
YoStarKR BattleStageName and EncounterOCR traslation missing item modifed


Constrat:
Syntax LGTM

PR #7804, opened 2023-12-27
chore: YoStarKR ocr text 

HX3N:
fix ocr text & add missing items


ABA2396:
Is there anything else that needs to be changed? I'm going to post a version now. Should I merge it?

PR #7846, opened 2023-12-31
perf: 优化刷理智-已开始战斗x次时临期理智药使用输出以增加可读性

status102:
改完之后好像又和自己冲了（

目前是 `当前理智: 当前量/上限, 药: 已使用的普通药(已使用的临期药)`

要不改成:n ? ,n ? {n} ? [n] ?

### DIFF
--- a/src/MaaWpfGui/Main/AsstProxy.cs
+++ b/src/MaaWpfGui/Main/AsstProxy.cs
@@ -833,7 +833,6 @@ namespace MaaWpfGui.Main
                             case "AnnihilationConfirm":
                                 StringBuilder missionStartLogBuilder = new();
                                 missionStartLogBuilder.AppendLine(LocalizationHelper.GetString("MissionStart") + $" {execTimes} {LocalizationHelper.GetString("UnitTime")}");
-                                var log = string.Empty;
                                 if (SanityReport.HasSanityReport)
                                 {
                                     missionStartLogBuilder.AppendFormat(LocalizationHelper.GetString("CurrentSanity"), SanityReport.Sanity[0], SanityReport.Sanity[1]);
@@ -853,7 +852,7 @@ namespace MaaWpfGui.Main
                                     missionStartLogBuilder.AppendFormat(LocalizationHelper.GetString("StoneUsedTimes"), StoneUsedTimes);
                                 }
 
-                                Instances.TaskQueueViewModel.AddLog(log.TrimEnd('\n').TrimEnd(' '), UiLogColor.Info);
+                                Instances.TaskQueueViewModel.AddLog(missionStartLogBuilder.ToString().TrimEnd('\n').TrimEnd(' '), UiLogColor.Info);
                                 break;
 
                             case "StoneConfirm":
--- a/src/MaaWpfGui/Main/AsstProxy.cs
+++ b/src/MaaWpfGui/Main/AsstProxy.cs
@@ -831,20 +831,26 @@ namespace MaaWpfGui.Main
                         {
                             case "StartButton2":
                             case "AnnihilationConfirm":
-                                var log = LocalizationHelper.GetString("MissionStart") + $" {execTimes} {LocalizationHelper.GetString("UnitTime")}\n";
+                                StringBuilder missionStartLogBuilder = new();
+                                missionStartLogBuilder.AppendLine(LocalizationHelper.GetString("MissionStart") + $" {execTimes} {LocalizationHelper.GetString("UnitTime")}");
+                                var log = string.Empty;
                                 if (SanityReport.HasSanityReport)
                                 {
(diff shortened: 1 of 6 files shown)


status102:
改完之后好像又和自己冲了（

目前是 `当前理智: 当前量/上限, 药: 已使用的普通药(已使用的临期药)`

src/MaaWpfGui/Main/AsstProxy.cs
```diff
@@ -839,7 +839,7 @@ private void ProcSubTaskStart(JObject details)
 
                                 if (MedicineUsedTimes > 0 || ExpiringMedicineUsedTimes > 0)
                                 {
-                                    log += $"{LocalizationHelper.GetString("MedicineUsedTimes")} {MedicineUsedTimes}" + (ExpiringMedicineUsedTimes > 0 ? $"({ExpiringMedicineUsedTimes})" : string.Empty) + "  ";
+                                    log += $"{LocalizationHelper.GetString("MedicineUsedTimes")} {MedicineUsedTimes}" + (ExpiringMedicineUsedTimes > 0 ? $"/{ExpiringMedicineUsedTimes}" : string.Empty) + "  ";
```
status102, line 842:
```suggestion
                                    log += $"{LocalizationHelper.GetString("MedicineUsedTimes")} {MedicineUsedTimes}" + (ExpiringMedicineUsedTimes > 0 ? $"<{ExpiringMedicineUsedTimes}>" : string.Empty) + "  ";
```

Constrat:
Do we have any space to add a:
Normal: `x` Expiring: `y`

status102:
> Do we have any space to add a:
> Normal: `x` Expiring: `y`

EN maybe ok. But there isn't enough space for ZH, because I try to make them in single line.

Constrat:
Do we have to stay in a single line? EN will definitely NOT stay in a single line. Unless with CN you can use just "one" symbol for each. 变: `x` 变: `y`

status102:
> Do we have to stay in a single line?

I try to make it. The smaller the output space, the better. Unless we can modify the output log. I once imagined an output similar to a console progress bar, but it was more troublesome to implement.
like this:
[ █ █ █ ..................... ]

image

Here the output about the number of battles and scores is collapsed into one line. However, if the score is relatively large, it will still wrap. I wonder if it will be optimized in the future.

Of course, this depends deeply on the language itself. I couldn't do same things in EN because of my bad English.

Constrat:
If it's difficult in chinese, in EN might be impossible. After working with both clients and OCR I noticed how much better kanji-based languages are better. They are usually much shorter buttons and sentences. Do what you can. At the end of they day it needs to be readable first, then it can "look cool".

status102:
> At the end of they day it needs to be readable first, then it can "look cool".

yah, you are right. And do you like this one?
```
Sanity: 0/135  Medicine: 5 / 2 expiring, Stone: 1
```
Or latter?
```
Sanity: 0/135  Medicine: 5 normal / 2 expiring, Stone: 1
```

status102:
> After working with both clients and OCR I noticed how much better kanji-based languages are better. 

It also causes more OCR faults. That's why I changed SSS stage ocr.

Constrat:
Latter is definitely more readable:
The 5 means only 5 used potions. Not of the 5 how many are expiring. In this case 7 potions were used correct?
Because we could change it to: Total / Expiring out of total maybe to clear it up a bit?

Constrat:
> > After working with both clients and OCR I noticed how much better kanji-based languages are better.
> 
> It also causes more OCR faults. That's why I changed SSS stage ocr.

Yeah, the fact the it always goes a new line (or that the resolution DPI Changes the rendering)

status102:
> Yeah, the fact the it always goes a new line (or that the resolution DPI Changes the rendering)

有个很奇怪的事情是，明明在相同的空间内，英文的复杂程度看上去比中文更小（英文只有26个字符）。但是MAA的使用介绍中对于外服却是1920x1080起步，国服只需要1280x720。

Constrat:
> > Yeah, the fact the it always goes a new line (or that the resolution DPI Changes the rendering)
> 
> One of the strange things is that in the same space, English seems less complex than Chinese (only 26 characters). However, in the introduction to the use of MAA, the foreign service starts at 1920x1080, and the national service only needs 1280x720.

Less complexity adds to an actual more complex OCR and or regex. I would have never guessed, but the Chinese OCR is far better than the English one (maybe because it was tweaked a lot more, it would make much more sense, as MAA started for the official Chinese client)

Constrat:
@status102 
image
uuuuh... This doesn't look quite right?

status102:
> @status102 image uuuuh... This doesn't look quite right?

My fault. Fixed now

PR #7859, opened 2024-01-01
chore: YoStarJP roguelike nextlevel ocr

Manicsteiner:
只测试了水月肉鸽的烧水模式，能正确退出，其他就但愿能用了


resource/global/YoStarJP/resource/tasks.json
```diff
@@ -5852,6 +5852,34 @@
...
+                "悪夢の幕開け",
+                "噩梦之始"
+            ],
+            [
+                "緋月の回廊",
+                "血月长廊"
+            ],
+            [
+                "渇欲の大広間",
+                "渴欲大厅"
```
Constrat, line 5879: Is the ocrreplace needed?
b72ac8d7d307e02c0cd2e285ab7b3d37e3c0a0f9 this was for EN (KR and TXWY also only use "text" and not ocrreplace)

resource/global/YoStarJP/resource/tasks.json
```diff
@@ -5869,6 +5962,38 @@
...
+                "不可侵の洞窟",
+                "失尘岩洞"
+            ],
+            [
+                "深海の密林",
+                "幽海丛林"
+            ],
+            [
+                "紺碧の揺籃",
+                "绀碧摇篮"
```
Constrat, line 5993: Same here.

Constrat:
I tried searching in MaaCore but `Roguelike@NextLevel` seems to get used only in the `Foldartal` tasks, which should only exist for the Sami I.S.
@Lancarus since you added these do you remember the use? For adaptation.

zzyyyl:
`Roguelike@NextLevel` 在 `Roguelike@Stages` 里，会被很多任务引用的

上面提到的 ocrReplace 方法估计也没问题，但是也许可以和 EN、txwy 等统一一下，直接修改 text 就可以了

Constrat:
> The ocrReplace method mentioned above is probably fine, but maybe it can be unified with EN, txwy, etc., and just modify the text directly

So what's the best approach? Ocrreplace or not?
https://github.com/MaaAssistantArknights/MaaAssistantArknights/commit/b72ac8d7d307e02c0cd2e285ab7b3d37e3c0a0f9 or #7859

zzyyyl:
> > The ocrReplace method mentioned above is probably fine, but maybe it can be unified with EN, txwy, etc., and just modify the text directly
> 
> So what's the best approach? Ocrreplace or not? [b72ac8d](https://github.com/MaaAssistantArknights/MaaAssistantArknights/commit/b72ac8d7d307e02c0cd2e285ab7b3d37e3c0a0f9) or #7859

I think `text` would be better.

Constrat:
This should be a bit better, and consistent with the other clients @zzyyyl @Manicsteiner

PR #7861, opened 2024-01-01
fix: encounter number in Mizuki IS

Constrat:
This was changed in #6102  4a6316c76a8db1e5ca101b14c53ec2ff429056d0 @Yumi0606 @Lancarus
This is the encounter for the 3rd ending. There should only be two options. Not three. In my case, MAA selected the "second" options out of 3 which resulted in choosing the 3rd ending which loses a lot of light.

### DIFF
--- a/resource/roguelike/Mizuki/encounter.json
+++ b/resource/roguelike/Mizuki/encounter.json
@@ -216,7 +216,7 @@
         },
         {
             "name": "开端",
-            "option_num": 3,
+            "option_num": 2,
             "choose": 2
         },
         {
--- a/resource/roguelike/Mizuki/encounter_for_deposit.json
+++ b/resource/roguelike/Mizuki/encounter_for_deposit.json
@@ -216,7 +216,7 @@
         },
         {
             "name": "开端",
-            "option_num": 3,
+            "option_num": 2,
             "choose": 2
         },
         {


Yumi0606:
image

Yumi0606:
From wiki.There is indeed a situation where the MAA chooses three endings and loses the light, which was originally intended to be modified in the previous update, but I have no time TwT.
Your modification seems to be wrong.Maybe it should be "Choose":3 and "option_num":3 ?

Constrat:
> image

Now It makes sense. This update is still not live.

Constrat:
> Your modification seems to be wrong.Maybe it should be "Choose":3 and "option_num":3 ?

image
As you can see in our version there's still only two options.

Pitiedwzr:
> > Your modification seems to be wrong.Maybe it should be "Choose":3 and "option_num":3 ?
>
> As you can see in our version there's still only two options.
>
I think you need to clear any ending to unlock the second option 'Caerula', for my game there are three options.
MuMuPlayer_TqsgN6MKuw

Constrat:
> > > Your modification seems to be wrong.Maybe it should be "Choose":3 and "option_num":3 ?
> > 
> > 
> > As you can see in our version there's still only two options.
> 
> I think you need to clear any ending to unlock the second option 'Caerula', for my game there are three options. MuMuPlayer_TqsgN6MKuw

Oh right, I forgot about that one. You need to get a clear to unlock the new "stuff"

Constrat:
My bad everyone

Constrat:
image
I got it.

PR #7894, opened 2024-01-03
chore: YoStarKR LE Stage navigation

HX3N:


Constrat:
@HX3N Check https://github.com/MaaAssistantArknights/MaaRelease/pulls for the hot update navigation

PR #7896, opened 2024-01-03
chore: YoStarKR LE stage navigation

HX3N:
revised regex


resource/global/YoStarKR/resource/tasks.json
```diff
@@ -51,6 +51,28 @@
...
+    },
+    "LEChapterToLE": {
+        "ocrReplace": [
+            [
+                ".?은노.?전당",
+                "붉은노을전당"
+            ],
+            [
+                ".*?전당",
+                "붉은노을전당"
```
Constrat, line 69: This is a bit redundant.
`.*?전당` will bypass `.?은노.?전당` because it's included in the regex.

Constrat, line 69:
```suggestion
            [
                ".*?전당",
                "붉은노을전당"
```

Constrat: This is a bit redundant.
.*?전당 will bypass .?은노.?전당 because it's included in the regex.

Constrat: No worries. We are for this (and we need KR player badly)

Constrat: It's usually better and suggested to create different PRs (for different purposes and changes) to allow easier reverting / debugging / maintaining.

PR #7959, opened 2024-01-08
refactor: 重构肉鸽投资流程，增加投资速度

status102:
image


src/MaaWpfGui/Res/Localizations/en-us.xaml
```diff
@@ -615,6 +615,7 @@ The video aspect ratio needs to be 16:9 without interference factors such as bla
     <system:String x:Key="EmergencyDps">Node: Emergency Operation</system:String>
     <system:String x:Key="DreadfulFoe">Node: Dreadful Foe</system:String>
     <system:String x:Key="RoguelikeEvent">Event:</system:String>
+    <system:String x:Key="RoguelikeInvestment">Invested {1}(+{0}), deposit: {2}</system:String>
```
Constrat, line 618: deposit is the current total amount of already invested ingots correct?
en i18n lgtm

Constrat (approved):

src/MaaWpfGui/Res/Localizations/en-us.xaml
status102: yes, it's an ocr result from here.
image

PR #8024, opened 2024-01-16
chore: YoStarKR ocr text update for SO LONG ADELE

HX3N:
add SL stage navigation and new character
include (#8018)


HX3N:
@Constrat

resource/global/YoStarKR/resource/tasks.json
```diff
@@ -51,24 +70,6 @@
...
-        "ocrReplace": [
-            [
-                ".*?전당",
-                "붉은노을전당"
-            ]
-        ],
-        "text": [
-            "붉은노을전당"
-        ]
-    },
```
Constrat, line 71: This should not be removed (we are not in MaaRelease)

HX3N: Oh, I see. I've restored it again.
I'm not sure, so I'm asking a question.
Do I need to update both MaaRelease and MaaAssistantArknights for Stage Navigation?

HX3N:
Oh, I see. Thank you for letting me know.

PR #8082, opened 2024-01-25
fix: 自动战斗-战斗列表的突袭关描述从`Adverse`修正为`Raid`

### LINKED ISSUE #8080: i18n: Challenge Mode / Adverse Environment / Hard Mode incosistency
### Reason

In most of the GUI localizations `Adverse` is used to describe the harder stage difficulty. This is not entirely accurate as the harder mode should be `Challenge`.

This issue became apparent and could create confusion considering the new main Story stages from Chapter 10 and after.

Chapter 10 introduced the Adverse Environment which allowed players to quickly play `Challenge` stages, but as of now, only the main story uses the "new" term `Adverse`, all sidestories and mini events use `Challenge`.

The term `Hard` is never used in-game specifically, but it can definitely be used in the GUI to explain both `Challenge` and `Adverse`.

Using the term `Adverse` in the Battle List of the Copilot tab is just wrong. As technically this is not an Adverse Environment (from the main story, but just a Challenge Mode of a Sidestory).
image

As far as I can understand zh_cn, the Official client uses `突袭` (translated as `raid`) for Challenge Mode. Is it also used for stages in Adverse Environment?

### Inconsistency:
In the Farming tab, the user is required to add `-Hard` in case it wants MAA to use `Adverse` environment (this only works for Chapter 10, 11, 12 etc, as chapter 9 uses the new Environment interface but doesn't have the option for `Adverse`)

### Suggestions:
1. Consistency > Clarity: Just use `Hard` for all menus and tasks, instead of `Adverse` in Copilot and `Hard` in farming.
2. Clarity > Consistency: considering that through `prts.plus` and Copilot we can clear pretty much every stage in the game (both `Adverse` and `Challenge` stages, the most logical solution would be to just switch the `Copilot` tab with a generic `Hard` (unless we can identify which stages are `Adverse` and which are `Challenge`, honestly a pain and a useless thing, as the user would need to know what stage is what. For this reason `Hard` seems the most logical answer.
Meanwhile in the `Farming` stage implementing `Adverse` as the manual entry of stages it mostly for farming stages and no `Challenge` stages are ever farmed / cleared automatically through farming.

Wow I just noticed how much I wrote. TL;DR just read the `Suggestions`

status102:
`(突袭)` for chinese
`(Raid)` for other languages

### DIFF
--- a/src/MaaWpfGui/Views/UI/CopilotView.xaml
+++ b/src/MaaWpfGui/Views/UI/CopilotView.xaml
@@ -197,6 +197,7 @@
                     x:Name="CopilotList"
                     Grid.Row="0"
                     Grid.ColumnSpan="4"
+                    HorizontalContentAlignment="Stretch"
                     dd:DragDrop.IsDragSource="{Binding Idle}"
                     dd:DragDrop.IsDropTarget="{Binding Idle}"
                     hc:BorderElement.CornerRadius="4,4,0,0"
@@ -210,18 +211,25 @@
                         <DataTemplate>
                             <Grid>
                                 <Grid.ColumnDefinitions>
-                                    <ColumnDefinition Width="150" />
                                     <ColumnDefinition />
+                                    <ColumnDefinition Width="Auto" />
+                                    <ColumnDefinition Width="Auto" />
                                 </Grid.ColumnDefinitions>
-                                <!--<maa:TextBlock  Text="{Binding ID}" />
-                            <maa:TextBlock  Text=". " />-->
                                 <CheckBox
                                     Grid.Column="0"
                                     Content="{c:Binding Name}"
+                                    Grid.IsSharedSizeScope="False"
                                     IsChecked="{Binding IsChecked}"
                                     IsHitTestVisible="{Binding ElementName=CopilotList, Path=DataContext.Idle}" />
-                                <Button
+                                <TextBlock
                                     Grid.Column="1"
+                                    MinWidth="35"
+                                    VerticalAlignment="Center"
+                                    Text="{DynamicResource CopilotTaskRaidStage}"
+                                    Visibility="{c:Binding IsRaid}" />
+                                <Button
+                                    Grid.Column="2"
+                                    MinWidth="30"
                                     hc:IconElement.Geometry="{StaticResource CloseGeometry}"
                                     BorderThickness="0"
                                     Command="{s:Action DeleteCopilotTask}"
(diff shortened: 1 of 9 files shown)


resource/tasks.json
```diff
@@ -718,12 +718,12 @@
...
         "next": [
-            "AdverseConfirm",
+            "RaidConfirm",
             "#self"
         ]
     },
-    "AdverseConfirm": {
-        "baseTask": "ChangeToAdverseDifficulty",
+    "RaidConfirm": {
+        "baseTask": "ChangeToRaidDifficulty",
```
Constrat, line 726: As I said before changing the image is not really ideal because Adverse is Adverse, challenge is challenge. You can't generalize it. 
`BattleStartAdverse`
BattleStartAdverse
`BattleStartRaid`
BattleStartRaid

status102: RaidDifficulty
image
NormalDifficulty
image

这里应该使用Raid的，一开始是看错`BattleStartRaid`和`BattleStartAdverse`了

status102:
`(突袭)` for chinese
`(Raid)` for other languages

src/MaaWpfGui/Views/UI/CopilotView.xaml
```diff
@@ -210,18 +211,25 @@
...
                                     <ColumnDefinition />
+                                    <ColumnDefinition Width="Auto" />
+                                    <ColumnDefinition Width="Auto" />
                                 </Grid.ColumnDefinitions>
-                                <!--<maa:TextBlock  Text="{Binding ID}" />
-                            <maa:TextBlock  Text=". " />-->
                                 <CheckBox
                                     Grid.Column="0"
                                     Content="{c:Binding Name}"
+                                    Grid.IsSharedSizeScope="False"
```
status102, line 221:
```suggestion
```

src/MaaWpfGui/Views/UI/CopilotView.xaml
```diff
@@ -210,18 +211,25 @@
                         <DataTemplate>
                             <Grid>
                                 <Grid.ColumnDefinitions>
-                                    <ColumnDefinition Width="150" />
                                     <ColumnDefinition />
+                                    <ColumnDefinition Width="Auto" />
+                                    <ColumnDefinition Width="Auto" />
```
status102, line 216:
```suggestion
                                    <ColumnDefinition />
                                    <ColumnDefinition />
```

PR #8450, opened 2024-03-02
chore: update resources for YostarKR

178619:
I have been so busy that I couldn't do much here but I can get back to work on this for some time and probably update some stuff along with the new roguelike theme in YostarKR (maybe)


Constrat:
Hey, glad to see you back :)

178619:
As far as I know it has not been working for a long time, so I assumed that the file is no longer used and deprecated. Also I actually had been using my own automated script to update it ever since the file got ids. I just didn't include it in the repo as I used to load the data from my local files.

PR #8487, opened 2024-03-05
chore: YoStarKR Sami roguelike templates and tasks

HX3N:
``` Sami@Roguelike@StageFerociousPresageEnter.png ``` 

The following template image is missing, just like YoStarJP. (I haven't encountered that event yet.)
I personally tested it and cleared it, but it's not certain. (OCR recognition issues)

Remove client: JP


HX3N:
Remove client: JP

PR #8587, opened 2024-03-12
perf: reduced IS stages according to new Auto Updater KR

HX3N:
Reference 58a2dc831cdbbad4b2cf72269290826f28429873

removed the spaces additionally in ocrReplace


Constrat:
Oh, it seems like for the KR there weren't any duplicates.

PR #8568, opened 2024-03-10
chore: YoStarKR update tasks

HX3N:
removed the ```squard``` and ```effect``` from ```RoguelikeTraderShoppingOcr``` (8a28f57ac00f8e4914052316872aa7cd823ede9a)

Also, modified the regex of some ocrReplace items that were not being recognized.

There was an issue with recognition in "巍峨银凇", but it has been resolved.
(Both regular space and nbsp were mixed and recognized)

```
[2024-03-10 19:52:06.703][TRC][Px2f180][Tx2f4c0] asst::WordOcr [{ 우뚝 솟은 은빛 서리: [ 0, 0, 473, 55 ], score: 0.896238 }] by OCR Rec , cost 100 ms
[2024-03-10 19:52:06.704][TRC][Px2f180][Tx2f4c0] Proceed [{ 巍峨银凇: [ 404, 453, 473, 55 ], score: 0.896238 }]
```

PR #8612, opened 2024-03-14
fix: Foldartal use for IS4 - overseas

Constrat:
+ Resource Update automaton

@HX3N 
Should be the same thing as the one mentioned in https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/8568#issuecomment-1997157055.
Before merging I want to make sure it doesn't break anything in the KR. (EN & JP have no changes so they will most likely follow KR and hoopefully work as well.)


HX3N:
> Before merging I want to make sure it doesn't break anything in the KR. (EN & JP have no changes so they will most likely follow KR and hoopefully work as well.)

It seems fine. I just double-checked, and it's working without any issues.

Constrat:
@HX3N It works. Seems like without the ocrreplace for the "usage" MAA internally couldn't understand the totems I guess. (strange).
Probably the Gain task was relegated only to identifying what was just received, and the Use was meant for everything else.

PR #8645, opened 2024-03-17
Feat: YostarEN SSS#3

Constrat:
~~I can't understand why the original author decided to give skill_usage: 1 to Penance and Typhon.
Penance dies rather easily without barriers.~~

~~@junyihan233 Sorry to bother, but I'm trying the SSS before merging and Penance would die even in Normal mode without activating her skill. (Same thing for Typhon, there is not enough DPS).
Unless there's a major reason that I'm completely missing I can't find a reason to not use the skills.~~

(Or I may be completely wrong, I just checked the docs, and 1 should activate the skill as many times as needed).
Ok I was completely wrong. Translating the names in the json completely broke the skill activations. My bad.


Constrat:
@HX3N @Manicsteiner You guys might be missing the ocrReplace for the Conductive Unit upgrade. 
ref:
- [竞技记录分析仪](https://prts.wiki/w/%E7%AB%9E%E6%8A%80%E8%AE%B0%E5%BD%95%E5%88%86%E6%9E%90%E4%BB%AA)
- [溢能回收转化器](https://prts.wiki/w/%E6%BA%A2%E8%83%BD%E5%9B%9E%E6%94%B6%E8%BD%AC%E5%8C%96%E5%99%A8)
- [战地援护无人机](https://prts.wiki/w/%E6%88%98%E5%9C%B0%E6%8F%B4%E6%8A%A4%E6%97%A0%E4%BA%BA%E6%9C%BA)

PR #8665, opened 2024-03-20
fix: YoStarKR ocrReplace

HX3N:
fixed character, encounter, foldartal


Constrat:
Just as a warning. After the Trader Shopping ocr incident where too many replaces were creating issues, if you can try to not add duplicates replaces. (I also used to it like this, to mantain the full name, in case it was needed and to debug when necessary).
A workaround is this:

```json
            [
                "Support Replenishment[ Station]*",
                "支援补给站"
            ],
```
instead of :
```json
            [
                "Support Replenishment",
                "Support Replenishment Station"
            ],
            [
                "Support Replenishment Station",
                "支援补给站"
            ],

```

resource/global/YoStarKR/resource/tasks.json
```diff
@@ -6359,14 +6363,27 @@
...
+                "트리마운츠여행사특별파견단"
+            ],
             [
                 "트리마운츠여행사특별파견단",
                 "特里蒙旅行社特派团"
             ],
+            [
+                "N6번공공이동.+",
+                "N6번공공이동스테이션"
+            ],
```
Constrat, line 6386: Just as a warning. After the Trader Shopping ocr incident where too many replaces were creating issues, if you can try to not add duplicates replaces. (I also used to it like this, to mantain the full name, in case it was needed and to debug when necessary).
A workaround is this:

```json
            [
                "Support Replenishment[ Station]*",
                "支援补给站"
            ],
```
instead of :
```json
            [
                "Support Replenishment",
                "Support Replenishment Station"
            ],
            [
                "Support Replenishment Station",
                "支援补给站"
            ],

```
This way, it's still completely readable, as to what the Original string should be, but doesn't create issues for MAA.

Theoretically it might create issues, where it's completely unredable and more chars that are not in the list are ocrr'ed, but the case of that happening should be minimal

HX3N: Unfortunately, KR cannot use "[]" because it breaks the regex. (crash)

The reason for changing it like this is that the word "스테이션" may or may not be recognized, "테" can be confused with "터" or "텨,".

To adhere to the above recommendations, it seems best for KR to change it as follows.

```json
            [
                "756번전파증폭스(테|터)이션",
                "在地公共信号放大站756号"
            ],
            [
                "트리마운츠여행사(특별파견단|.+)",
                "特里蒙旅行社特派团"
            ],
            [
                "N6번공공이동스(테|터)이션",
                "在地公共移动站N6号"
            ],
```

Constrat: That should also work :). The important is trying to reduce the amount of replaces MAA needs to do. (And preferably keep the original string intact (for easier debugging / code navigation)

PR #8737, opened 2024-03-30
ci: overrides the output encoding to utf-8 for python

wangl-cc:
Fixes [#8728](https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/8728#issuecomment-2028194915)

把这个先合了试试？


wangl-cc:
#8739

PR #8768, opened 2024-04-03
chore: YoStarJP ocr fix

Manicsteiner:


resource/global/YoStarJP/resource/tasks.json
```diff
@@ -2812,6 +2824,14 @@
...
                 "アーモンド",
                 "杏仁"
+            ],
+            [
+                "Friston-$",
+                "Friston-3"
+            ],
+            [
+                "U-[oO]ficial",
+                "U-Official"
```
ABA2396, line 2834: image
参考这个吧

Constrat:
In EN fortunately there no issues regarding this but.
For KR and JP isn't the CN official regex enough?

resource/global/YoStarJP/resource/tasks.json
```diff
@@ -2812,6 +2824,14 @@
...
                 "アーモンド",
                 "杏仁"
+            ],
+            [
+                "Friston-$",
+                "Friston-3"
+            ],
+            [
+                "[Uu]-[O0o]f{1,2}icial",
+                "U-Official"
```
Constrat, line 2834:
```suggestion
```

Constrat: Shouldn't this work as there is already the CharsReplaceOcr in `tasks.json`

Constrat:
Just repeated the commit changing the name to allow merging.
(I think I finally understood how to fix the name thing through reset -> force push

PR #8769, opened 2024-04-03
fix: 尝试修复肉鸽导航 oversea clients

Manicsteiner:
#8764 消除8764对其他服务器的影响
已在日服测试，可用
Request test for EN, KR @Constrat @HX3N


Constrat (approved):

Constrat:
It doesn't look good, but for scalability and ease of access the use of multiple same screenshots is approved this time.
When the update rolls for Global we can just remove everything from this PR and it should work (hopefully)

PR #8774, opened 2024-04-03
perf: removing redundant regex for CharsNameOcrReplace

Constrat:
Unless I'm missing something, shouldn't the Official CN task inheritance do all the work?
ref: d1a1010e53a5ee11fe3d4cddc6c50539d9e9d8b9


Constrat:
So in the case of long arrays, it won't add them but just consider the one corresponding to the client?

ABA2396:
When you display a declared field, its value is no longer related to the original.

PR #8943, opened 2024-04-30
fix: Mac编译脚本自动移除之前的build文件夹

AndersonALAC:
auto remove previous builds upon build
or it could cause errors when rebuilding

### DIFF
--- a/tools/build_macos_universal.zsh
+++ b/tools/build_macos_universal.zsh
@@ -4,6 +4,8 @@ basedir=$(dirname "$0")/..
 pushd ${basedir}
 
 build_arch() {
+    // remove previous artifacts
+    rm -rf build-$1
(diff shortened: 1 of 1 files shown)


wangl-cc:
If we remove the `build-*` directories, all compile cache will be removed. However, this script seems to be designed to use cache, because it will always re-download the dependencies. So this PR is OK for me.

tools/build_macos_universal.zsh
```diff
@@ -4,6 +4,8 @@ basedir=$(dirname "$0")/..
 pushd ${basedir}
 
 build_arch() {
+    // remove previous artifacts
```
wangl-cc, line 7: Use `#` instead of `//` for comment in zsh.

wangl-cc (changes requested):

PR #9068, opened 2024-05-11
chore: YoStarJP ocr fix

### LINKED ISSUE #9065: 日服自動戰鬥無法識別及部屬夜鶯、煌、阿三位幹員；自動編隊無法識別閃靈
### 在提问之前...

- [X] 我理解 Issue 是用于反馈和解决问题的，而非吐槽评论区，将尽可能提供更多信息帮助问题解决
- [X] 我填写了简短且清晰明确的标题，以便开发者在翻阅 issue 列表时能快速确定大致问题。而不是“一个建议”、“卡住了”等
- [X] 我已查看最新测试版本的更新内容，并未提及该 bug 已被修复的情况
- [X] 我已检查了置顶议题（pinned issue），确认我的问题未被提及

### 问题描述

H13-3 maa://29078 手動編隊進入戰鬥後無法識別及部屬夜鶯，以致後續行動均失效 05-11 15:57:09.247
斑點 悖論模擬 maa://36682 無法部屬幹員阿，以致後續行動均失效 05-11 16:06:36.871
H13-2 maa://29075 自動編隊無法識別閃靈。無法部屬煌，以致後續行動均失效  05-11 16:08:34.593

### Version

Resource Version: ツヴィリングトゥルムの黄金
UI Version: v5.2.4-alpha.1.d059.g4dd3f0cba
Core Version: v5.2.4-alpha.1.d059.g4dd3f0cba

### 日志文件

[asst.log](https://github.com/MaaAssistantArknights/MaaAssistantArknights/files/15282126/asst.log)
[gui.log](https://github.com/MaaAssistantArknights/MaaAssistantArknights/files/15282128/gui.log)

### 截图

_No response_

### 还有别的吗？

BS 5
win 10

Manicsteiner:
close #9065


resource/global/YoStarJP/resource/tasks.json
```diff
@@ -259,6 +259,10 @@
                 "ブレイズ",
                 "煌"
             ],
+            [
+                "ブレイス",
+                "煌"
```
Constrat, line 264:
```suggestion
                "ブレイ[ズス]",
                "煌"
            ],
```
Why not

Constrat (commented):
I'm noticing a lot of regexes that can be squashed in one.
Remember that after a total max of regexes it seems MAA will crash (this was happening in the TraderShoppingOcr)

resource/global/YoStarJP/resource/tasks.json
Manicsteiner: The Japanese version has an exclusive "equivalence_classes" in ocr_config.json, which sometimes oddly doesn't work when the "one of" ([ ]) usage appears in same ocr_replace rule, which I have started to avoid.

I have seen MAA crash due to ocr replace only one time. My impression is that in the tested ocr replacement rules, the replaced text can also be matched by the replacement rules, resulting in repeated ocr replacement and eventual crash.
I noticed that there are some replacement rules in the English and Korean versions that continue to replace non-Chinese content with non-Chinese content, as well as some global rules that replace non-character content into empty. Unexpectedly circular substitutions occur in these rules, may be the truth behind running out of OCR times.
Of course, I will try to reduce unnecessary substitution rules in the future to avoid encountering the real upper limit of substitutions.

Constrat: What was happening with the TraderShopping seemed to have been resolved by reducing it under ~600 ocr replacement cases.
As long we try to reduce what's possible, without going crazy, everything's good.
With EN I'm close to 550 replaces, while JP seems to be almost 700 (and nothing has happened yet luckily).

PR #9083, opened 2024-05-12
perf: auto start optimization + MAC fix 

Constrat:
I have no way to test for MAC

### DIFF
--- a/src/MaaCore/Task/Miscellaneous/StartGameTaskPlugin.cpp
+++ b/src/MaaCore/Task/Miscellaneous/StartGameTaskPlugin.cpp
@@ -4,59 +4,50 @@
 
 using namespace asst;
 
+bool StartGameTaskPlugin::start_game_with_retries(size_t pipe_data_size_limit, bool newer_android)
+    const
+{
+    int extra_runs = 0;
+    for (int i = 0; i < 30; ++i) {
+        if (!ctrler()->start_game(m_client_type)) {
+            return false;
+        }
+
+        if (ctrler()->get_pipe_data_size() > pipe_data_size_limit) {
+            if (newer_android || ++extra_runs > 3) {
+                return true;
+            }
+        }
+
+        sleep(1500);
+    }
+
+    return false;
+}
+
 bool StartGameTaskPlugin::_run()
 {
     if (m_client_type.empty()) {
         return false;
     }
+
     // check for MAC / iOS
     if (ctrler()->get_controller_type() == ControllerType::MacPlayTools) {
         return ctrler()->start_game(m_client_type);
     }
-    else {
-        // check for android version, because it leads to different magic values
-        // https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/8966#issuecomment-2094369694
-        if (ctrler()->get_version() > 8) {
-            do {
-                if (!ctrler()->start_game(m_client_type)) {
-                    return false;
-                }
 
-                // https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/8961#issue-2277568882
-                // 167: magic value needs to be >164 but <172 (as that's max)
-                if (ctrler()->get_pipe_data_size() > 167) {
-                    return true;
-                }
+    // check for android version, because it leads to different magic values
+    // >8:  167: magic value needs to be >164 but <172 (as that's max)
+    // <=8: 145: needs to be >85 but <153 (as that's max)
+    // https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/8966#issuecomment-2094369694
+    // https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/8961#issue-2277568882
+    int pipe_data_size_limit = (ctrler()->get_version() > 8) ? 167 : 145;
 
-                sleep(1500);
+    // In Android 9 and above, there's a way to check if the game started correctly
+    // so no need for limited tries
+    bool newer_android = (ctrler()->get_version() > 8);
 
-            } while (true);
-        }
-        else {
-            do {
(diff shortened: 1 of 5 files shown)


ABA2396:
This feels a bit strange. How about adding an enum for platform, and then PlayTools just skips over(Maybe a later version of PlayTools will do something else)

Constrat:
tbf, I have no idea how to do that XD. I managed to copy paste the existing AdbController to implement the version and the pipe_data_size.

ABA2396:
m_controller_type?

Constrat:
The best solution would be to just:
```c++
if(Android) {
	    if (ctrler()->get_version() > 8) {
	        do {
	            if (!ctrler()->start_game(m_client_type)) {
	                return false;
	            }
	            if (ctrler()->get_pipe_data_size() > 167) {
	                return true;
	            }
	            sleep(1500);
	        } while (true);
	    }
	    else {
	        do {
	            if (!ctrler()->start_game(m_client_type)) {
	                return false;
	            }
	            if (ctrler()->get_pipe_data_size() > 145) {
	                for (auto _ = 3; _--;) {
	                    sleep(1500);
	                    if (!ctrler()->start_game(m_client_type)) {
	                        return false;
	                    }
	                }
	                return true;
	            }
	            sleep(1500);
	        } while (true);
	    }
} else {
	if (!ctrler()->start_game(m_client_type)) {
		return false;
	}
}
```

Constrat:
> m_controller_type?

Just to make sure, the only one that works with PlayTools is just MacPlayTools, I should be able to use that yes?

ABA2396:
image
image
Honestly it doesn't feel like the touch mode and the platform is a strong tie either, but for now it's fine
cc @status102 @MistEO

Constrat:
Oh I like this. I was directly using the TouchModes.

ABA2396:
cc @hguandl

Constrat:
On Android (Mumu12) it worked like before.

src/MaaCore/Task/Miscellaneous/StartGameTaskPlugin.cpp
```diff
@@ -9,48 +9,53 @@ bool StartGameTaskPlugin::_run()
     if (m_client_type.empty()) {
         return false;
     }
+    // check for MAC / iOS
+    if (ctrler()->get_controller_type() == ControllerType::MacPlayTools) {
+        return ctrler()->start_game(m_client_type);
+    }
+    else {
```
ABA2396, line 16: The else can be omitted here, right? That's too many layers of indentation.

src/MaaCore/Task/Miscellaneous/StartGameTaskPlugin.cpp
```diff
@@ -9,48 +9,53 @@ bool StartGameTaskPlugin::_run()
...
 
-            // https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/8961#issue-2277568882
-            // 167: magic value needs to be >164 but <172 (as that's max)
-            if (ctrler()->get_pipe_data_size() > 167) {
-                return true;
-            }
+                sleep(1500);
 
-            sleep(1500);
+            } while (true);
```
ABA2396, line 33: I think it's better to add a maximum number of retries.

src/MaaCore/Task/Miscellaneous/StartGameTaskPlugin.cpp
Constrat: Oh you're right, there's a return either way

src/MaaCore/Task/Miscellaneous/StartGameTaskPlugin.cpp
Constrat: No idea what a good number would be.
Perhaps 50 just to make completely sure.

src/MaaCore/Task/Miscellaneous/StartGameTaskPlugin.cpp
```diff
@@ -4,54 +4,50 @@
 
 using namespace asst;
 
+bool StartGameTaskPlugin::start_game_with_retries(size_t pipe_data_size_limit, bool newer_android)
+    const
+{
+    int extra_runs = 0;
+    for (int i = 0; i < 30; ++i) {
```
status102, line 11: I think we need `need_exit()` in any loop

Constrat: I can easily add it with a if and a return, but is automatic? Or we do need to add more logic to tell him when the need_exit is needed?

ABA2396: > I can easily add it with a if and a return, but is automatic? Or we do need to add more logic to tell him when the need_exit is needed?
> 
> Tbh, at that point if we know a condition where we can stop the loop, we can just put it in the loop condition, instead of using need_exit()

The core will be notified when the ui hits stop, and need_exit() will return the information.

Constrat: Oh, interesting. Well considering it's a stop, we can add it exactly as the first step of the loop, to stop it as fast as possible.

Constrat:
[7ab466a](https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/9083/commits/7ab466a17beda8e2d539e4d23bbf707f505ab525) @status102 @ABA2396 need_exit() added, I've checked other files and this seems to be the implementation

Constrat:
I'll check one last time the stop function.

ABA2396:
合了发个公测版吧，mac没有内测的更新，之前更新了公测版的应该都用不了了

PR #9434, opened 2024-06-21
perf: 优化仓库识别结果展示，支持多语言显示

ABA2396:
baf12f45ab9e3806d5040103deed8055
990d697f72329449be9b4be76880b4e0

cc @Constrat due to longer item names in English, we can only fit two per line compared to three in Chinese. Any thoughts on this?

### DIFF
--- a/src/MaaWpfGui/Helper/ItemListHelper.cs
+++ b/src/MaaWpfGui/Helper/ItemListHelper.cs
@@ -15,6 +15,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Text.Json;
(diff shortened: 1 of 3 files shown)


Constrat:
Personally if it works, it works. Looking at the UI, making the font smaller to allow for a 3rd column is excessive.
It'll require more scrolling but I don't really have more ideas on this, unless maybe moving the number counter under the name? But even then it's just going around the problem.

On another note, does increasing the size of the display window allow for more columns with longer name? (Just looking at the size of Polymerization Preparation, it doesn't seem possible).

Most important thing is that it works :)

Constrat (approved):

PR #9538, opened 2024-07-06
chore: update python API

EvATive7:
The following interfaces are supported via this pr：
- set_connection_extras
- set_static_option

### DIFF
--- a/src/Python/asst/asst.py
+++ b/src/Python/asst/asst.py
@@ -137,6 +137,23 @@ class Asst:
         return Asst.__lib.AsstConnect(self.__ptr,
                                       adb_path.encode('utf-8'), address.encode('utf-8'), config.encode('utf-8'))
 
+    def get_image(self, size: int) -> bytes | None:
+        """
+        获取上次截图
+        :params:
+            ``size``:  图像字节数, 如 1280*720*3
+
+        : return: 成功时图像的字节; 失败时 None
+        """
+        buffer_type = ctypes.c_byte * size
+        buffer = buffer_type()
+        buffer.value = b'\000' * size
+        if (got := Asst.__lib.AsstGetImage(self.__ptr, buffer, size)) \
+                and got > 0:
+            return bytes(buffer)
+        else:
+            return None
+
     def set_connection_extras(name: str, extras: JSON):
         """
         连接模拟器端的Extras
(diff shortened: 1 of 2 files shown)


aur3l14no:
Maybe add `get_image` as well?

```py
    def get_image(self, size) -> bytes | None:
        """
        获取上次截图

        : return: Image in bytes or None on failure
        """
        buffer = ctypes.create_string_buffer(b'\000' * (size - 1))
        if (size := Asst.__lib.AsstGetImage(self.__ptr, buffer, len(buffer))) \
            and size > 0:
            return bytes(buffer)
        else:
            return None

...

        Asst.__lib.AsstGetImage.restype = ctypes.c_uint64
        Asst.__lib.AsstGetImage.argtypes = (
            ctypes.c_void_p, ctypes.c_void_p, ctypes.c_uint64)
```

aur3l14no:
Being picky here :D

1. The docstring of `set_static_option` are not consistent with params.
2. Maybe add return value to `set_connection_extras`?
3. Maybe add `get_image` as well?

```py
    def get_image(self, size) -> bytes | None:
        """
        获取上次截图
        :params:
            ``size``:  图像字节数, 如 1280*720*3

        : return: 成功时图像的字节; 失败时 None
        """
        buffer_type = ctypes.c_byte * size
        buffer = buffer_type()
        buffer.value = b'\000' * size
        if (got := Asst.__lib.AsstGetImage(self.__ptr, buffer, size)) \
            and got > 0:
            return bytes(buffer)
        else:
            return None

// ...

        Asst.__lib.AsstGetImage.restype = ctypes.c_uint64
        Asst.__lib.AsstGetImage.argtypes = (
            ctypes.c_void_p, ctypes.c_void_p, ctypes.c_uint64)
```

Example usage: `asst.get_image(1280*720*3)`

EvATive7:
Thanks for your reply.

1. Got it. I will fix it.
2. MAACore won't return anything for calling `AsstSetConnectionExtras`. This is its prototype:
```cpp
void AsstSetConnectionExtras(const char* name, const char* extras)
```
3. Thanks for your code, I will add it.

SherkeyXD (approved):

bigbanglwb:
can i use `set_connection_extras ` to enabel mumu extras ? which means enable `mumu 截图增强模式`. Is there any documents for this api ? Thank u

PR #9539, opened 2024-07-06
feat: YoStarEN resolution warning on client switch

Constrat:
What do you think
image


Constrat:
I would like to add a link to explain why that's needed but considering the docs files are all over the place I'll think about it.

Constrat:
While I was working in the debug XAML part of the project I noticed this. Which makes entering the settings tab with some delay, as it seems there are some errors behind it. Everything still works, but the delay is noticeable.

System.Windows.Data Error: 4 : Cannot find source for binding with reference 'RelativeSource FindAncestor, AncestorType='HandyControl.Controls.TagContainer', AncestorLevel='1''. BindingExpression:Path=(0); DataItem=null; target element is 'Tag' (Name=''); target property is 'ShowCloseButton' (type 'Boolean')

docs/用户手册/新手上路.md
```diff
@@ -37,7 +37,7 @@ winget install Microsoft.DotNet.DesktopRuntime.8
 
 6. 正确设置模拟器分辨率
 
-模拟器分辨率应为 `16:9` 比例，最低为 `1280x720`；对于国际服（美服）玩家，最低为 `1920x1080`。
+模拟器分辨率应为 `16:9` 比例，最低为 `1280x720`；唯一支持的分辨率是 `1920x1080`。
```
ABA2396, line 40:
```suggestion
模拟器分辨率应为 `16:9` 比例，最低为 `1280x720`；对于国际服（美服）玩家，唯一支持的分辨率是 `1920x1080`。
```

ABA2396:
Yostar is a lowercase s
image

Constrat:
... I had to change it because all the code was written with a Upper case. The first try I did was with `_clientType is "YostarEN"`. But it wasn't working.

@HX3N @Manicsteiner So from now on, please use Yostar in the commits, instead of YoStar. I'll see if I can refactor everything without creating issues.

ABA2396:
It looks like this is a legacy issue. Let's modify the interface to display everything in lowercase, but we shouldn't change anything in the protocol.

Constrat:
I'll do a merge from dev. Because the changelog and other things have been modified

PR #9542, opened 2024-07-06
fix: switch account failed when MuMuEmulator12Extras is enabled

EvATive7:


Constrat:
Screenshot timing is too fast yes?

EvATive7:
> Screenshot timing is too fast yes?

yes. After opening the account selection panel, should wait for a while before taking screenshot, otherwise MumuExtra screenshots will be too fast and the account list will not be captured, and will slide down the account list and miss the account supposed to select

PR #9552, opened 2024-07-07
docs: 增加关于肉鸽任务刷探索范式功能的说明

Alan-Charred:


SherkeyXD (approved):

Constrat:
@SherkeyXD Regarding #9551 I think this is already a problem. I can understand that the english documentation won't be really used as the majority of the userbase is still cn, but this goes to increase the discrepancy in the docs languages.

I noticed the webplate feature request implementation. That should definitely help, but I hope for future PRs that go to modify / add docs to wait just a bit, or to get some i18n, acknowledgements regarding i18n.

Not necessarily translating, but just creating the empty file and maybe creating a stupid simple file with: TO TRANSLATE and hopefully when I have some time I can start translating (ML unfortunately) but still something.

SherkeyXD:
Good idea. We may add a check or guideline later.

PR #9555, opened 2024-07-07
refactor: 肉鸽资源加载逻辑及目录结构重构

SherkeyXD:
Todos:
~~+ [ ] autocopilot 也重构一下~~
+ [x] encounter 采用类似 tasks.json 的覆盖模式，各个模式的事件配置覆盖到 `default.json` 上面
+ [x] 肉鸽插件加载主题化（待定）
+ [x] Wait for #9561


Constrat:
Since we are adding in all global languages the tip on top of the webpage, should we disable this? 
image
It's a bit annoying as it's asked every time.

SherkeyXD:
> Since we are adding in all global languages the tip on top of the webpage, should we disable this? 

We are now considering to change the `Translation Needed` tip into a sign meaning that the doc isn't translated at all or hasn't been updated for a very, very long time.

As for the popup, we can make it show only once.

PR #9569, opened 2024-07-08
chore: combox behaves as normal combox without IsEditable

Constrat:
Unfortunately `IsEditable` is used as a "way" to cover the button for deletion. But that adds some other issues such as not being able to open the combobox from anywhere, but only from the arrow dropdown.

Removing `IsEditable` shows the button which is a bit ugly but reimplements the full functionalities of the combobox 
image
image

I'm not very knowledgeable in XAML but maybe we can color the button border to the foreground color to mask it a bit more.


ABA2396:
IsEditable is actually for the user to copy the current configuration name for naming new configurations or for writing call scripts, and can only be opened by clicking on the drop-down arrow, which is actually what I expected (x)

Constrat:
I honestly even forgot what this was about. If it's working as intended I can close :)

PR #9866, opened 2024-07-22
feat: close inactive issues

Constrat:
ref: https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues/9853#issuecomment-2243230911

`exempt-issue-labels`  can be used to skip (for example) `MAA Team`  issues, or any issue that has a specific label, plus the custom made label `keep-open`, in case we want to use that.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -24,7 +24,7 @@ jobs:
           stale-issue-message: |
             "This issue has been inactive for a prolonged period and will be closed automatically in 3 days."
             "该问题已长时间处于闲置状态，3 天后将自动关闭。"
-          exempt-issue-labels: "keep-open"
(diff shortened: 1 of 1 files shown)


zzyyyl:
It's annoying :laughing:

Constrat:
We'll to reach 1k issues 😢

zzyyyl (approved):

Rbqwow:
I wonder which time zone github bot uses 👀

Constrat:
UTC -> GMT 0

Constrat:
I've been rate limited. Going to have to wait some time. For now I'll just increase the cron.

dantmnf:
👀 Well, it's a good way to tell people "stop opening issues here".

PR #9927, opened 2024-07-26
feat: first implementation

Constrat:
I'm currently not sure if I shuold double check if there were real png changes. Currently this workflow should run only when changes are made to these paths:
```yaml
      - "resource/global/txwy/resource/template/**"
      - "resource/global/YostarEN/resource/template/**"
      - "resource/global/YostarJP/resource/template/**"
      - "resource/global/YostarKR/resource/template/**"
      - "resource/template/**"
      - "docs/.vuepress/public/image/**"
      - "website/apps/web/public/**"
      - "website/apps/web/src/assets/links/**"
```

Which are pretty much the locations of the pngs currently. Meaning that if a change happens there it's 95% a png change or addition.
There's the last 5% that makes me not sure. Meaning the optimizer could run when no optimization is actually needed. Nothing bad happens, but it's just wasted time. Which compared to the git diff might be less.

### DIFF
--- a/.github/workflows/optimize-templates.yml
+++ b/.github/workflows/optimize-templates.yml
@@ -2,6 +2,8 @@ name: Optimize PNG Templates
 
 on:
   push:
+    branches:
+      - "dev"    
     paths:
       - "resource/global/txwy/resource/template/**"
       - "resource/global/YostarEN/resource/template/**"
--- a/.github/workflows/optimize-templates.yml
+++ b/.github/workflows/optimize-templates.yml
@@ -16,7 +16,7 @@ on:
       commit_message:
         description: "Commit Message"
         type: string
-        required: false    
+        required: false
 
 jobs:
   check-and-optimize-png:
@@ -31,7 +31,7 @@ jobs:
       - name: Run optimization script
         run: |
           python3 optimize-templates.py
-        
+
       - name: Commit changes
         id: commit_changes
         run: |
@@ -39,7 +39,7 @@ jobs:
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
 
           git add .
-          
+
           $commit_msg = "${{ github.event.inputs.commit_message }}"
           if (-not [string]::IsNullOrWhiteSpace($commit_msg)) {
             $commit_msg = $commit_msg.Trim()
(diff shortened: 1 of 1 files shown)


Constrat:
Technically the git diff could be used to give the paths to the optimize.py but I don't even think it's worth it.

dantmnf:
Then we need to store both the old and new version in the git repository 👀

Maybe we can do this in a pre-merge rebasing (or just for release packaging)

Constrat:
Regarding the pre release step I'm definitely with you. I like all the prettify and template optimization but I feel like they are too much bothering especially in the `dev`  branch where we push a lot and except for nightlies we shouldn't really care a lot about.

Constrat:
> Then we need to store both the old and new version in the git repository 👀

@dantmnf I couldn't quite catch what you meant here.

Anyway my opinion, is that all this stuff is call, but it's a pain in the ass XD. Especially the prettyfying step as it serves almost no purpose, since the default json formatting is not that bad.

As I said previsously, the png optimization actually makes a lot more sense, but we should do that only in master in m opinion. Because as you can see everytime a png is modified we can another duplicate with the optimized template that overwrites the original one. These steps should be done directly in master (automated with workflow), that way master only has optimized and dev has not optimized that allows us devs and contributors to not really care about all that.

dantmnf:
Well, my point is "avoid unnecessarily bloating the repository". Quite developer-centric : )

> I couldn't quite catch what you meant here.

I mean optimized PNG and the "unoptimized" one here. By putting the optimized PNG in a new commit, Git allows you to checkout the previous commit that contains the unoptimized one. Therefore we devs will need to download ~2x the size for each new template, which I think is opposed to the sense of optimization.

> that way master only has optimized and dev has not optimized

This way suffers from the same issue, as `git clone` will fetch all visible refs (branches) in the repository. This is also the reason why I prefer pushing WIP changes to my own fork rather than opening a new branch here.

Constrat:
Oh, I know see what you meant, but unfortunately to "fix" this it would require to just ignore the optimization (and remove pre-commit.ci which I would gladly accept), or optimize directly in the push to dev.

(The correct solution would be to just push everything to a fork like you said and then merge the PR to dev, but this is much more work to fork, PR and merge)

My stand has always been: I don't give a shit about these "beautifying steps" considering they just add commits and (personally) unecessary requirements and changes.

PR #9928, opened 2024-07-26
ci: optimize templates perf and fixes

Constrat:


Constrat:
Ops, during testing I changed the identity lmao

PR #9984, opened 2024-07-30
perf: 优化萨卡兹肉鸽ew精二优先级

Daydreamer114:


ABA2396:
~~The original phrase "いいよ！こいよ！" is a line from a Japanese adult video where the character Beast Senpai says it during a climax scene.~~

However, it can also be understood as a play on the Japanese words "良い世，来いよ," which translates to "Come, beautiful world."

Constrat:
Wow, such lore

PR #9995, opened 2024-07-31
chore: roguelike navigation for YostarJP

Manicsteiner:
Revert #8769 for YostarJP only


Constrat:
@Manicsteiner The change was for all clients? I'm thinking the same issue should happen with all globals yes?

Manicsteiner:
Yes, this change has been made to all Yostar servers, but you need to check out the new text section by yourself. Whats else, score of `Roguelike@IntegratedStrategies.png` is only around 0.75, and I am not sure why.

Constrat:
Well EN is still behind as the update should drop with Shu Event yes so in 8 hours

Constrat:
> Yes, this change has been made to all Yostar servers, but you need to check out the new text section by yourself. Whats else, score of `Roguelike@IntegratedStrategies.png` is only around 0.75, and I am not sure why.

It's so small that the minimal change reduces the score by quite a lot?

Manicsteiner:
Its score dropped to around 0.55 at first today, but when I tried to replace it with a new one and swap it back to compare, it became 0.99. It's weird. Maybe it's due to caching.
[asst-rg-0.55.log](https://github.com/user-attachments/files/16448416/asst-rg-0.55.log)
[asst.log](https://github.com/user-attachments/files/16448419/asst.log)
just search `Roguelike@IntegratedStrategies.png`

无标题
Tested with red dot, it shouldn't matter

PR #9996, opened 2024-07-31
chore: StartToVisit ROI for Yostar servers

Manicsteiner:
They pre-installed this new interface


Constrat:
They already pushed it? wow

Constrat:
@HX3N Yeah I guess so, since we removed the custom task name, the official one should be without texts

PR #10023, opened 2024-08-01
chore(style): strings instead of ints for training room time output

Constrat:
Before:
image

After:
image

Reading times as x:x:x gives me a stroke.

(Why no space between `[Name] Skill` 😭 )

@HX3N no more stroke when reading

### DIFF
--- a/src/MaaCore/Task/Infrast/InfrastTrainingTask.cpp
+++ b/src/MaaCore/Task/Infrast/InfrastTrainingTask.cpp
@@ -153,10 +153,13 @@ bool asst::InfrastTrainingTask::time_left_analyze(cv::Mat image)
         }
         std::string str_time = match.str();
 
-        if (!utils::chars_to_number(str_time, time_left[i])) {
-            Log.error(__FUNCTION__, "chars_to_number failed");
-            return false;
+        // Ensure the time string has two characters
+        if (str_time.length() == 1) {
+            str_time = "0" + str_time;
         }
+
+        // Store the time as a string with leading zeros
+        time_left[i] = str_time;
     }
 
     return true;
--- a/src/MaaCore/Task/Infrast/InfrastTrainingTask.h
+++ b/src/MaaCore/Task/Infrast/InfrastTrainingTask.h
@@ -25,7 +25,7 @@ namespace asst
         static int skill_index_from_rect(const Rect& r);
 
         int m_level;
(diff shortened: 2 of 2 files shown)


src/MaaCore/Task/Infrast/InfrastTrainingTask.cpp
```diff
@@ -153,10 +153,13 @@ bool asst::InfrastTrainingTask::time_left_analyze(cv::Mat image)
         }
         std::string str_time = match.str();
 
-        if (!utils::chars_to_number(str_time, time_left[i])) {
-            Log.error(__FUNCTION__, "chars_to_number failed");
-            return false;
+        // Ensure the time string has two characters
+        if (str_time.length() == 1) {
+            str_time = "0" + str_time;
         }
```
Constrat, line 159:
```suggestion
		std::ostringstream oss;
		oss << std::setw(2) << std::setfill('0') << str_time;
		str_time = oss.str();
```
This could be used but it seems stupidly more complicated

ABA2396:
Sorry, I directly made the changes when I saw the issue and didn't notice this PR. I don't think storing time as a string is a good idea. This display issue should be handled by the UI.

ABA2396:
image
just like this

status102:
or send a number of seconds, instead of h, m, s

Constrat:
@ABA2396 
Instead of opening an issue I'll just use this as discussion:
image
How about adding a space in between? 
```cpp
   // Before
   $"[{operatorName}]{skillName}\n" +

   // After
   $"[{operatorName}] {skillName}\n" +
```

I know that the in-game output does not have the space, but especially for English, not having it on an interface feels wrong. Let me know your opinions, in case you don't like for all client, I can just add it to English only if possible

status102:
> @ABA2396我不会提出问题，而是将其用作讨论： image中间加一个空格怎么样？
> 
> ```c++
>    // Before
>    $"[{operatorName}]{skillName}\n" +
> 
>    // After
>    $"[{operatorName}] {skillName}\n" +
> ```
> 
> 我知道游戏中的输出没有空间，但特别是对于英语来说，界面上没有它感觉不对。让我知道您的意见，如果您不喜欢所有客户，我只能在可能的情况下将其添加为英文

@ABA2396 says it's not a matter. If you like, just add it

PR #10040, opened 2024-08-02
chore: training room styled output depending on language

Constrat:
This is the more complicated solution to https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/10023#issuecomment-2265033067.

This way the change is only for EN. In case the change is liked on all clients, I can just 
```cpp
$"[{operatorName}] {skillName}\n" +
```


Constrat:
EDIT: resolved https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/10023#issuecomment-2265068333

PR #10058, opened 2024-08-03
chore: 更新issue - bug report模板

Daydreamer114:


Constrat (approved):
Looks good, not that it's gonna help reduce the dumb issues as people are just going to ignore it XD

PR #10079, opened 2024-08-04
feat: 允许选择WPFGUI中的干员名称显示语言

Manicsteiner:
允许选择“干员名称显示语言”为“跟随MAA”或“跟随游戏客户端”。
这个应该不会需要更多选项吧（）
先糊一坨，睡了（）

已知问题：
 - 所有相关内容都需要重启客户端才能生效

TODO：
 - 将“切换语言后重启MAA”引入此处选项修改后任务
 - Other languages
 - 更多显示干员名称的区域，如 #10067 提到的招募小工具界面
 - 拆分重置 #9457 

待讨论：
 - 更多的专有名词

图片


Constrat:
Big effort required, especially for the copilot part. 
Easiest way to implement without bothering the entire front-end, back-end of prts.plus and the logic for the operators extraction from the json would be to just iterate a search for the correct operator name against a glossary. (The `docs/glossary.md` we had before could have been fairly useful), but fortunately we have `battle_data.json` that collects operator, traps and operator gadgets. 
The problem (I think) is that the search would be fairly heavy (?). As it's required to search the operator first, and then print the correct language directly (if not in zh-cn already).

```json
        "char_002_amiya": {
            "name": "阿米娅",
            "name_en": "Amiya",
            "name_jp": "アーミヤ",
            "name_kr": "아미야",
            "name_tw": "阿米婭",
            "position": "RANGED",
            "profession": "CASTER",
            "rangeId": ["3-6", "3-1", "3-1"],
            "rarity": 5
        },

```

Manicsteiner:
先就这么多功能吧，再塞太多又要等N久

ABA2396:
我感觉我穿越了，因为我之前好像翻译过导能原件之类的东西，但现在我又找不到了）

Manicsteiner:
> 我感觉我穿越了，因为我之前好像翻译过导能原件之类的东西，但现在我又找不到了）

就是提到的 #9457，正在拆分重置

ABA2396:
好好好

ABA2396:
这个已经整完了吗？如果好了的话解决一下冲突直接合了吧

Manicsteiner:
似乎出现了一处错误，等我检查一下
图片

Manicsteiner:
commit https://github.com/MaaAssistantArknights/MaaAssistantArknights/commit/80c34a07e1191b36c36ab826a1b29f27e6e3b2f5 改动导致了这个问题，与本pr无关，但也是需要改正的问题

ABA2396:
> commit [80c34a0](https://github.com/MaaAssistantArknights/MaaAssistantArknights/commit/80c34a07e1191b36c36ab826a1b29f27e6e3b2f5) 改动导致了这个问题，与本pr无关，但也是需要改正的问题

直接把这个 SetAndNotify 改回来好了

ABA2396 (approved):

Constrat:
@Manicsteiner 
image
Missing skills?

Manicsteiner:
> #### Btw I like the Follow MAA or follow language, but this way a user can't select what language he wants for operators, for example I play EN but I can't check CN currently.

My initial idea was that users should always set the MAA language option as their native language, just allowing them to choose whether to translate these proper nouns when playing non-native language game clients. 
Now it seems that it is necessary to consider the situation where non Chinese users have to use Chinese or English when integrating with other tools. 
However, allowing users to choose any language here will only turn it into a second language option, and a further discussion of the extent is needed. 
Here is an example of the 'Always display of Chinese' option [feat_opername_r1](https://github.com/Manicsteiner/MaaAssistantArknights/tree/feat_opername_r1)

Constrat:
What about a secret config setting? Like the nightly releases?

PR #10127, opened 2024-08-07
feat: YostarKR Reclamation2

HX3N:


Constrat:
Did you notice any ROI that need fixing / tweaking?

HX3N:
> Did you notice any ROI that need fixing / tweaking?

I just tested the normal mode, and it seems that we don't need to tweak the roi.

Constrat:
> I just tested the normal mode, and it seems that we don't need to tweak the roi.

I am ready for the worst possible string display in the international :)

HX3N:
KR also needs to be modified 😭

Constrat:
Yeah I think all Yostar servers.
They changed the schedule compared to Official and Bilibili. But my question remains valid.
Since reclamation is a perma mode now, why would they need to shove it in your face? When they could have just put the banner, instead of replacing and switching everything.

PR #10128, opened 2024-08-07
chore: YostarJP HS navigation update

Manicsteiner:


Constrat:
Of course Yostar changes schedule and we get fucked over

Issue #9992, opened 2024-07-31
[Feature] Reclamation Algorithm for Yostar servers


Manicsteiner:
### 说说你遇到的问题？

<details>
<summary>Coming soon</summary>

![mcQ2u-6z6yK2sT3cSsg-fy 1](https://github.com/user-attachments/assets/bf6a5f22-e28c-4e05-9843-63b7545cd10c)

</details>

**TODO**:

- [x] Revert #8769 (have to be manually now), fix roguelike navigation
- [x] Unlock Reclamation Algorithm#2 for EN, KR and JP (txwy needs to wait)
- [x] Templates and texts for Reclamation Algorithm
- [x] Endless tests

Constrat:
I hate this fucking mode

HX3N:
Pre-preparation

I think we only need to add the image `Reclamation2SkipDays1.png`. Except for that, they are simply shapes

![Reclamation2SkipDays1](https://github.com/user-attachments/assets/959dc92d-1286-4bf4-bc14-55e004a35f53)

<details><summary>Arrange tasks in order (with image)</summary>

    "Reclamation2Begin": {
        "algorithm": "OcrDetect",
        "action": "ClickSelf",
        "text": ["开", "始", "演", "算"],
        "roi": [844, 569, 436, 151],
        "next": ["#self", "#back"]
    },
![image](https://github.com/user-attachments/assets/d7bc93ec-577d-4b58-a8c8-d496dd3e3a07)

    "Reclamation2BeginPlotEnter": {
        "algorithm": "OcrDetect",
        "action": "ClickSelf",
        "text": ["众人", "汇聚", "后舍"],
        "roi": [633, 239, 275, 156],
        "next": ["#self", "Reclamation2BeginPlotLoading@Reclamation2EnterStage"]
    },
![image](https://github.com/user-attachments/assets/aaeec443-7d8d-4e85-8276-f4d083ed89a8)

    "Reclamation2LeaveStageConfirm": {
        "algorithm": "OcrDetect",
        "action": "ClickSelf",
        "text": ["确认", "离开"],
        "roi": [871, 407, 238, 146],
        "next": ["#back"]
    },
![image](https://github.com/user-attachments/assets/76dad085-49dc-4ff4-b572-f81d455a54d1)

    "Reclamation2CopiousCoppice": {
        "Doc": "进入丰饶灌木林，基地上方支线",
        "algorithm": "OcrDetect",
[...]

PR #10143, opened 2024-08-08
feat: YostarEN Reclamation2

zayn7lie:


HX3N:
You can refer to [this comment](https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues/9992#issuecomment-2267951824) for ocr text in other mode

Constrat (approved):
Just did a run with it. Everything worked flawlessly.
@zayn7lie Quick question. Are you using 1920x1080?
In case you're not, somehow everything still worked on my 1920x1080 environment (which I suggest as some texts are stupidly small sometimes)

HX3N:
Wait a second, it seems like `Reclamation2ExReadArchiveSelectDayConfirm` is missing

Constrat:
I was about to check the Ex one now. I need to understand how the item craft are interpreted in the core. Directly ocred? Like with IS starting operator?

HX3N:
> I was about to check the Ex one now. I need to understand how the item craft are interpreted in the core. Directly ocred? Like with IS starting operator?

It recognizes the equipment list using OCR, selects the identified equipment, and proceeds with the crafting. (If the OCR recognition is inaccurate, it seems like it doesn't work)

Constrat:
Yeah I guess exactly like the IS one. You need the client language to fill it. No `CharsNameOcrReplace` is done

zayn7lie:
Wait a sec. I am still debugging with that.

Since mode 1 requires enough level and days in that, I need some time to reach the requirement for translation.

Constrat:
Yeah I noticed that something is needed to use the Ex correctly :P

zayn7lie:
> Just did a run with it. Everything worked flawlessly. @zayn7lie Quick question. Are you using 1920x1080? In case you're not, somehow everything still worked on my 1920x1080 environment (which I suggest as some texts are stupidly small sometimes)

Nope. I am using 1280x720. I think it is because of OCR could also handle with some scaling, rotating or so.

At the same time, WordOCR seems to do nothing much with resolution lol.

Constrat:
The issue is when \n are present. If you check this you will understand why I've been only using 1920 for the past year and a half: 
- https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues/5235

zayn7lie:
> The issue is when `\n` are present. If you check this you will understand why I've been only using 1920 for the past year and a half:
> 
>     * [Resolution Inconsistency: AutoSquad Copilot Operator's name Detection in YostarEN #5235](https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues/5235)
> 
> 
> Luckily this time the wording seems identical in both 720 and 1080 but that's usually the opposite.

I think there might be one way to handle with that, just separate the terms into words?

Constrat:
Most alters use the same words and usually when there's a \n the and the text is small the OCR does not consider the two words apart and mixes them leading to a unusable result.

zayn7lie:
> Most alters use the same words and usually when there's a \n the and the text is small the OCR does not consider the two words apart and mixes them leading to a unusable result.

NP, I will also handle with that later

Constrat:
Remember that too many ocrreplace will crash MAA as it already happened in the JP client with the Trader in IS.

Constrat:
image
Nvm I see why. Probably still haven't unlocked something aye

HX3N:
image

Constrat:
"Reclamation2ExReadArchiveEnter": {
    "algorithm": "OcrDetect",
    "action": "ClickSelf",
    "text": ["读取"],
    "roi": [1030, 548, 170, 124],
    "sub": ["Reclamation2ExReadArchiveSelectDay", "Reclamation2ExReadArchiveSelectDayConfirm"]
},

This is 

"Reclamation2ExReadArchiveEnter": {
    "text": ["Load Save"],
},

zayn7lie:
> image
> 
> should unlock the shop. It will probably open at base level 3

If it's a good idea to crop the middle of them?

Constrat:
I don't think the shop is needed in any of the actions for manufacturing. Increasing the ROI should not be a problem.

resource/global/YoStarEN/resource/tasks.json
```diff
@@ -2559,12 +2559,29 @@
...
         "text": ["Craft"]
     },
+    "Reclamation2ExManufactureConfirmAmount": {
+        "algorithm": "OcrDetect",
+        "Doc": "如果刚好用完的话可能会卡住，这个之后再处理了",
+        "text": ["99"],
+        "roi": [720, 442, 175, 166],
+        "sub": ["Reclamation2ExManufactureConfirm"],
+        "next": ["Reclamation2ExManufacture"]
+    },
```
Constrat, line 2569: This can be removed

zayn7lie:
> I don't think the shop is needed in any of the actions for manufacturing. Increasing the ROI should not be a problem.

Fine. The potential problem of different resolution might only happen on "Are you sure you want /n to leave?". For wordOCR and cropping it only scan "Are you sure you" so it will not cause a big problem lol. But anyway let me also handle with that.

zayn7lie:
NP. Feel free to have a check~

HX3N:
> I don't think the shop is needed in any of the actions for manufacturing. Increasing the ROI should not be a problem.

It seems to be because the task needs to be started when the game is already quite progressed. It wouldn't matter if the reclamation is initiated as soon as it is recognized, but wouldn't there be an issue when trying to load the saved survival days later?

Constrat:
No file changes are with ROI so let's leave it as is, without trying to center the ROI to make it work for both ways.

@HX3N I noticed you are missing two tasks on the korean?
    "Reclamation2ExBegin": {
        "text": ["Resume"]
    },
    
    and
    
    
  "Reclamation2ExReadArchiveConfirm": {
      "text": ["Confirm"]
  },

zayn7lie:
> > I don't think the shop is needed in any of the actions for manufacturing. Increasing the ROI should not be a problem.
> 
> It seems to be because the task needs to be started when the game is already quite progressed. It wouldn't matter if the reclamation is initiated as soon as it is recognized, but wouldn't there be an issue when trying to load the saved survival days later? (The Ex task involves repeating the process of crafting > loading saved survival days)

Do you mean the feature of "Archive Days"? I have done with "Archive Days" (or "Save" in game). Check [#10143](https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/10143/commits/c770d1579540dbd642f9109e3efd66d4343d6ef2)

zayn7lie:
> No file changes are with ROI so let's leave it as is, without trying to center the ROI to make it work for both ways.
> 
> @HX3N I noticed you are missing two tasks on the korean?
> 
> ```
>     "Reclamation2ExBegin": {
>         "text": ["Resume"]
>     },
> ```
> 
> and
> 
> ```
>   "Reclamation2ExReadArchiveConfirm": {
>       "text": ["Confirm"]
>   },
> ```

Day 4 is enough for "Save" feature. You could check with those ocr quickly~

Constrat:
Anyway we can merge. As most features are done.
We can always fix them later. To also clearup the discussion.

@zayn7lie Thank you very much :)

Constrat:
@HX3N I would like to push a nightly. Do you want to add the missing tasks and run the nightly yourself? Or do I just run it now?

HX3N:
> @HX3N I would like to push a nightly. Do you want to add the missing tasks and run the nightly yourself? Or do I just run it now?

do update. I have encountered a new problem, so it might take a while.

dontrunmk:
is RA2 currently available on the stable version?

HX3N:
> is RA2 currently available on the stable version?

Not possible in the stable version yet. It is currently only available in the Nightly version.

PR #10151, opened 2024-08-08
chore: convert workflow input values to int

Constrat:


Constrat:
nvm. I'm not built for this

Constrat:
Oh this seems to work

Constrat:
Now it's working correctly. Tested with:
- default values 
    - web 
    - mobile app
- custom value 
    - web 
    - mobile app

ABA2396:
I tried expr and it doesn't handle non-ints, I'll see what I can do about that.

Constrat:
Expr is not good. Printf seems to work but it's a huge instructions. In the end regexing seems good :)

Constrat:
And with this it should be good to merge (squash as there's too many test commits)

ABA2396:
image
It looks good; I like it.
By the way, where do I access the action on the phone? I couldn’t find it after searching for a while.

ABA2396:
Well, it looks like I didn't update it XD

Constrat:
:O

PR #10155, opened 2024-08-08
docs: add minor tip in CN doc and sync EN doc.

zkywalker230061:


docs/en-us/develop/development.md
```diff
@@ -85,7 +85,36 @@ Welcome to the [GitHub Pull Request Tutorial](./pr-tutorial.md) that anyone can
...
+1. Ensure that you have Python and Node environments on your computer.
+
+2. Execute the following command in the root directory of the project:
+
+    ```bash
+    pip install pre-commit
+    pre-commit install
+    ```
+
+If pre commit still cannot be used after pip install, please check if the pip installation path has been added to the PATH.
```
Constrat, line 113:
```suggestion
If pre-commit still cannot be used after pip install, please check if the pip installation path has been added to the PATH.
```

Constrat (approved):

PR #10174, opened 2024-08-09
chore: add int convertion to release-ota

Constrat:
Same as 
- #10151

for release-ota.yml

## Summary by CodeRabbit

- **新功能**
	- 改进了GitHub Actions工作流，增强了发布限制的输入解析功能，确保输入值为整数并进行确认回显。


Constrat:
Lots of notifications from this guy though!

AnnAngela:
@Constrat It seems like this AI only supports **1** natural language at same time. So we can not have both Chinese and English summary.

PR #10205, opened 2024-08-11
i18n: YostarEN Reclamation2QuitToMainPage

zkywalker230061:
Reclamation2QuitToMainPage: Back
Also some wired format tool seems to be running, shouldn't be a problem. (?)


zkywalker230061:
Done in another commit.

PR #10208, opened 2024-08-11
fix: ClickStageName pattern extraction for EX stages

Constrat:
In many cases, after first clearing a stage with a normal and challenge mode, the OCR will pickup a `I` or `J` from the "completion" logo,  which prevents the task from continuing and results in a task error.
image

```py
[2024-08-11 17:51:37.271][TRC][Px41a4][Tx0104] asst::CharOcr [...... { JHS-EX-2: [ 117, 316, 109, 32 ], score: 0.906266 } ...... { IHS-EX-3: [ 124, 513, 101, 32 ], score: 0.963113 }
```

This regex resolves the issue, but currently only for EX stages, as the problem hasn't been identified in other stages yet. This is primarily because only EX stages feature both a normal and a challenge mode.

The change is fairly simple. It utilizes` std::regex_replace` ability to use backreference substitution. (I didn't know it was possible directly in the JSON, but after checking how the values are passed from JSON to MaaCore and they remain simple strings it stupid easy to implement)

### DIFF
--- a/resource/tasks.json
+++ b/resource/tasks.json
@@ -298,7 +298,10 @@
         "isAscii": true,
         "text": [],
         "next": ["ClickedCorrectStageOrSwipe", "#self"],
-        "ocrReplace": [["-\\]", "-1"]]
+        "ocrReplace": [
+            ["-\\]", "-1"],
+            ["^.*([A-Z]{2}-EX-\\d+)", "$1"]
+        ]
     },
     "StageNavigationSlowlySwipeLeft": {
         "baseTask": "ChapterSlowlySwipeToTheLeft",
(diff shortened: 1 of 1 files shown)


Constrat:
As far as I remember, EX stages can only have 2 capital letters in front for events.

Constrat:
I don't understand the JP smoke testing error. It seems template related?

ABA2396:
You deleted the tasks Reclamation2SkipDaysAwardEndTask and Reclamation2SkipDaysBeginNewDayEndTask, which caused the Yostar* to be unable to override these tasks. I think you might have made an error during the rebase.

resource/tasks.json
```diff
@@ -298,7 +298,10 @@
         "isAscii": true,
         "text": [],
         "next": ["ClickedCorrectStageOrSwipe", "#self"],
-        "ocrReplace": [["-\\]", "-1"]]
+        "ocrReplace": [
+            ["-\\]", "-1"],
+            ["^.*([A-Z]{2}-EX-\\d+)", "$1"]
+        ]
```
ABA2396, line 304: I think that's the only modification. Everything else has been reverted.

Constrat:
Oh, how did this happen lol.

resource/tasks.json
Constrat: If you think it's good, I'll just commit to dev this :)

Constrat:
Acutally I'll just make a new one.

PR #10211, opened 2024-08-11
chore: added regex for EX stages

Constrat:
- #10208 but updated to `dev`


ABA2396 (approved):
In the regular expression, `^` specifies matching from the beginning of the string, but since `^` is immediately followed by `.*` (which matches any character), it renders `^` meaningless. Therefore, `^` can be removed without affecting the match result.

PR #10215, opened 2024-08-11
refactor: 整理生息演算<沙中之火>的任务流

Alan-Charred:


Constrat:
Why the need for the change?

Alan-Charred:
> Why the need for the change?

To ensure that in the long run we keep a more organised task-flow where reclamation algorithm tasks are named in the same way as the integrated strategy tasks... I know the changes I have made are messy. Apologies. On the other hand, considering that Fire-Within-the-Sand is no longer available in any of the servers, I don't think I am harming anything. Deleting these could also be an option... anyway I will come up with refined tasks for SandTales ASAP.

Besides refining code & resources for reclamation algorithm, I would like to suggest that we stick to some good practices of tasks naming like <Theme>@<Mode>@<MainUse>@<Additional name which directs to a base task where you can use #back to play some task magic>... Someone should provide a guideline :D

Oh if you have any good ideas on this, please let me know.

Cheers,

Constrat:
Thank you for the answer, I was in no way disregarding the idea, refactoring is good (sometimes).
I just wanted to make sure that the current RA2 iteration doesn't use ***ANY*** tasks from the original Fire within the sand.

zzyyyl:
有空可以改一下tasksorter，把生稀盐酸的排序也写进去

Alan-Charred:
> Thank you for the answer, I was in no way disregarding the idea, refactoring is good (sometimes). I just wanted to make sure that the current RA2 iteration doesn't use _**ANY**_ tasks from the original Fire within the sand.
> 

Yes that's also what I want to make sure, and I believe it does be the case.

> By the way, the official English translation is "Fire Within the Sand," so perhaps the tasks and templates should be named "FireSand"? It does have a nicer ring to it.

Aha, this is relevant to the grammar order in different languages. In Chinese grammar "Fire Within the Sand" will be translate to something like "The Sand's Fire". I also want to stick to the official name "Fire Within the Sand", but FireSand sound like focusing on Sand instead of Fire... maybe I am wrong. Well let's deleting the sand for now, and as long as we keep a good naming of tasks, renaming them in the future is just a replace `^FireReclamation` -> `FireSandReclamation`. I am also open to "FWTS" btw.

Alan-Charred:
> 有空可以改一下tasksorter，把你重构完的生稀盐酸的排序也写进去

好勒～

Constrat:
Technically, to minimize size and weight in our workflows and for the users, we could consider "archiving" the entire logic. Deletion is not advisable since we may need it again in the future. However, since they are currently not in use, we could refactor with an eye towards optimization, rather than merely renaming.

Alan-Charred:
> Technically, to minimize size and weight in our workflows and for the users, we could consider "archiving" the entire logic. Deletion is not advisable since we may need it again in the future. However, since they are currently not in use, we could refactor with an eye towards optimization, rather than merely renaming.
> 
> In my mind: if it's not being currently used and won't be used in the meantime, I will just remove it. Considering there's quite a lot of templates and some tasks. The only issue that I can think about is the MaaCore. Since the tasks are usually hardcoded, what happens if the Core calls a non-existing task in the json?

Totally agreed.
Actually there is a lot of optimisation to do, like removing redundancy, organise templates with subfolders -- given the absence of name conflict.
Several days ago I finished a tool/RoguelikeRecruitmentTool which provides visualisation , validation and automatic organisation (removing redundant fields & sort the fields).
Some other MAA team members is also developing similar tools for tasks, with which we may able to double check the dependency of tasks, whether certain tasks is called directly in cpp files, and prone certain will-no-longer-be-used branches from this task forest.

The Task.get function will return a nullptr, sometimes it will just be ignored, and sometimes it will block the entire task, depending on where the task is. We should definitely find a way to keep track of the tasks called directly in MaaCore.

Constrat:
Yeah the project is getting too big compared to the tools we have avaiable to "manage" it.
Lots of template as outdated and are not being used (event templates for example) which should all be archived. (Never delete anything, even with git history control, it's better to transfer to non build directories.

PR #10283, opened 2024-08-18
chore: git blame added styling commits

Constrat:
`git config blame.ignoreRevsFile .git-blame-ignore-revs` in cli is required to allow this to work in VSCode (and for Git Blame extension in case anyone is using it)

image
They will not disappear from the VSCode git timeline, which is good :)

ref to commits:
d84da5d01a872540d520bc87408d4524e07cfa39
546bfc399cdd140a9812f44b5e6bc89fccfda5fa
f0bdccdf0da7fac29f475951f0f43d3ab56a7ffe
7a758bce844bc2778f04a789195efbd17681dac7
5f3730b76785b1168125eb3b5edbed399722f8f9
bc3ad67a02cb3b8e154c06955cf17091c4265000


SherkeyXD (approved):

SherkeyXD:
Is there any way to add styling commits automatically?

Constrat:
We could use a super specific commit name/identifier and try to do something via workflow with a "simple" shell script (or maybe there's an already made GitHub action for this but I need to research it first).

I really don't have other solutions/ideas

PR #10322, opened 2024-08-21
i18n: KR GpuDeprecated translation

HX3N:


HX3N:
JP is also missing @Manicsteiner

PR #10335, opened 2024-08-22
feat: GetLocalizedNames for Infrast and Copilot output

Constrat:
Simple change that implements GetLocalizedNames for infrast output.
(Didn't think it was this easy, so maybe issues arise from this?)

It's basically a cheap copy of what @Manicsteiner has done for CopilotView.

_Would like to require approval first since C# is not my main_

Output example:
Infrast:
image
Copilot:
image
^ how about rethinking the: `Current step: Deploy operator`. The `Current step:` seems very redundant


Constrat:
Ok I never noticed the DataHelper class lol... Pretty powerful ngl, it's implementing all the things I was asking a year ago. Need to start closing issues :)

Manicsteiner (approved):

Constrat:
Nothing comes to mind regarding other fields that need localization.
I'll discover them using MAA.

PR #10424, opened 2024-09-02
feat: date format language based

Constrat:
c479834d38822ed43c8d2f2df3e3d7ba1154256e

### DIFF
--- a/.github/ISSUE_TEMPLATE/cn-bug-report.yaml
+++ b/.github/ISSUE_TEMPLATE/cn-bug-report.yaml
@@ -32,6 +32,9 @@ body:
         Resource Version:
         UI Version:
         Core Version:
+        或
+        Resource Version:
+        MAA Version:
     validations:
       required: true
   - type: textarea
--- a/.github/ISSUE_TEMPLATE/cn-mumu-report.yaml
+++ b/.github/ISSUE_TEMPLATE/cn-mumu-report.yaml
@@ -38,6 +38,10 @@ body:
         UI Version: 
         Core Version: 
         MuMu 版本号:
+        或
+        Resource Version:
+        MAA Version:
+        MuMu 版本号:
     validations:
       required: true
   - type: textarea
--- a/.github/ISSUE_TEMPLATE/en-bug-report.yaml
+++ b/.github/ISSUE_TEMPLATE/en-bug-report.yaml
@@ -45,6 +45,9 @@ body:
         Resource Version:
         UI Version:
         Core Version:
+        OR
+        Resource Version:
+        MAA Version:
     validations:
       required: true
   - type: textarea
--- a/MAA.sln.DotSettings
+++ b/MAA.sln.DotSettings
@@ -52,6 +52,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Ehem/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=esource/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Favourite/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=filetime/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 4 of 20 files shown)


ABA2396:
我个人感觉中文这样显示怪怪的，不过英语确实太长了看不到，可能放前面会好一点。
这个判断条件可能也有问题，资源名称和默认语言没有关系，而是和 client 选择有关，或者做成每个服务器的分开显示。
看看别人有没有什么看法。

I personally feel that it looks a bit odd in Chinese like this, but the English text is indeed too long and gets cut off, so maybe placing it at the front would be better.
The conditional logic might also be problematic—resource names aren't related to the default language but rather to the client's choice. Or perhaps it would be better to display them separately for each server.
Let's see if anyone else has any thoughts.

cc @MaaAssistantArknights/global-support-team @Rbqwow @status102

Rbqwow:
我觉得应该把日期独立出来单独做个叫“资源日期”的版本号
~~或者全部用 `-` 分割~~ 感觉还是不好看 但是我又想到了一个非常好的主意
`泰拉饭#0902`

ABA2396:
但是英语的活动太长了）
你也不想 What the Firelight Casts - Retrospection#0902 吧

Constrat:
#MMdd honestly looks ugly as fuck XD. It seems a github issue.
Anyway, a date is not related to the client but to the language personally.
If a chinese user uses YostarEN the date should be displayed with the Asiatic standard (yy.MM.dd).
If an english user uses Bilibili (I do) the date should be displayed with the English standard (dd.MM.yy)

Client usage has nothing to do with date standards

status102:
日期在前面看着怪怪的，而且年份可以去掉吧

Constrat:
The date at the start is related to the version itself instead of the resource. It doesn't make sense logically but

ABA2396:
In fact, this date is not related to the version, and related to the resource

Rbqwow:
我觉得我那个仿id风格挺好的）
正好井四位数

status102:
尝试换行，但是高度不会变

image

status102:
QF3L0022U{5(@M@`5{ILHTM
M{3X%PCR2~ A8VSC576QWRC

Give me a bit more time to make it great again x)

Rbqwow:
> 我有一个想法
> 把资源版本日期和软件版本日期都拆开来
> 这边塞5条
> 资源版本，资源版本日期，界面版本，核心版本，软件版本日期

—— uye aka. ABA2396

status102:
image

不管怎么说，搞定了

Constrat:
@Rbqwow The abbreviation is good I guess, but the value are automated from the ResourceUpdater. We don't manually modify the event names (we could but it's a lot of work, and they would get replaced the next time anyway)

ABA2396:
其实我想这样
```
资源版本 迷宫饭
资源日期 2024.09.02
UI 版本 5.6.0
core 版本 5.6.0
构建时间 2024.09.05
```
资源日期和软件版本日期分开显示，这样不容易有误解（b站已经有114514个人问了）

Constrat:
Displaying separately looks very good. Technically another solution could be implementing the date inside the version id / number but we are already using commits for nightlies, so it wouldn't make sense.

ABA2396:
> Displaying separately looks very good. Technically another solution could be implementing the date inside the version id / number but we are already using commits for nightlies, so it wouldn't make sense.

This is mainly used to remind the user, for example, the SS was released in 24.9.2, but you are using the 24.8.15 version, so it may not be used

ABA2396:
cc @status102 翻译还没写，明天在这个基础上改改吧，今天先休息了

ABA2396:
image

Constrat:
Wow, from 1 file change to 18 XD

ABA2396:
image
image

CultureInfo.CurrentCulture 认为 en 的 format 应该是 MM/dd/yyyy

Constrat:
:( that's the american standard, absolute garbage putting the month before the day

ABA2396:
image
Well, it's not a great idea, but you can use it to change the

ABA2396:
不管了，开合！

Constrat:
Hype !

PR #10487, opened 2024-09-05
chore: YostarJP CR new operators ocr

Manicsteiner:
> [!warning]
存在未确认的链接，请谨慎访问。
There are unconfirmed links, please visit with caution.

新干员和几个遗留问题
Doc的文字识别难以置信的出现了西里尔字母的o，而且可以稳定复现，只能先放到 equivalence_classes 里面了
图片
[asst - RU.log](https://github.com/user-attachments/files/16887604/asst.-.RU.log)


Daydreamer114 (approved):

resource/global/YoStarJP/resource/ocr_config.json
```diff
@@ -22,7 +22,7 @@
         ["へ", "ベ", "ペ", "ヘ", "べ", "ぺ"],
         ["ホ", "ボ", "ポ"],
         ["り", "リ"],
-        ["o", "O"],
+        ["o", "O", "о"],
```
Constrat, line 25: This is crazy OCR type of precision. 
When it's not needed the OCRer will find everything it doesn't need to

Manicsteiner:
> Since they are english names I just hope the english OCRer doesn't bother.

This may be because resource\global\YoStarJP\resource\PaddleOCR\rec\key.txt contains some Cyrillic letters and some Greek letters. I'm not sure if this list is the valid range or limited range of the OCR model output, but fortunately the EN version of this list does not include these kinds of letters.

PR #9943, opened 2024-07-27
feat: ResourceUpdater based on CMAKE

Constrat:
I don't know what I'm doing but it worked (somehow, kinda)

The objective of this PR is to switch the runner of the res-update-game.yml to allow it to run on a full ubuntu runner to hopefully increase speed.

As you can see this is my very first time using CMake and I have no idea what I'm doing. I pretty much just copied what was in the "main" CMakeLists.txt and tried to adapt it to make it work for the ResourceUpdater.sln. Some small codechanges were necessary for the `main.cpp` itself.

Basically, everything works except something a bit strange that doesn't happen on the Windows Runner. 

<details>
<summary>This:</summary>

image

</details>

For some reason it's also extracting these that come from the levels.json and they are the Co-op stages.
```json
    "name": "规则教学-常规项目",
    "code": "IG-TR-1",
    "levelId": "activities\\act2vmulti\\level_act2vmulti_tr01",
```
Why they get extracted here and not in windows? I have no idea. The stages are present in the `overview.json` so I'm actually thinking that the Windows runner is not extracting them, anyway, this should not really create any issues as the co-op stages are well, 1. Finished 2. MAA couldn't do them (unless jk))

I'm asking the big guys that have used CMake more than me to have a look and if necessary (very likely do whatever changes they feel like, as long as it works of course).

PR #10493, opened 2024-09-05
perf: resource updater validator in native powershell

Constrat:
> [!warning]
存在未确认的链接，请谨慎访问。
There are unconfirmed links, please visit with caution.

I'm a "purist" and probably have [OCD](https://en.wikipedia.org/wiki/Obsessive%E2%80%93compulsive_disorder).
I love optimizing stuff especially in a CI/CD environment (even though the runner is supposedly with unlimited minutes), but with the implementation of the node.js step, it takes a WHOPPING (joke) 15 seconds more.
We run the workflow 3 times an hour meaning 72 times a day. That's a whole **18 MINUTES** _wasted_.

So, this is the Powershell equivalent of https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/45e8548b077701164efa4ccdbfd0e56b4bc22f46/tools/ResourceUpdateValidator/index.mjs

I'm beginning testing now and hopefully it takes less than 15 seconds to run, in my local machine it's basically instant but it wouldn't be the first time the GH runner is ages slower compared to.


Constrat:
Logically the functioning _should_ be identical, the only small diff is the outputting as Powershell is fairly limited when outputting JSON-like structures: hashtables arrays etc, so I manually made it look like that. 

Honestly the JSON structure for output looks kind of bad as it doesn't feel clean but it's just outputting in the end (and I'll probably be the only one looking at those anyway XD)

Constrat:
Already posed the question in #10453 but I will ask again since it's still relevant:
Looking at the logic, in case the only file that is changed is the version.json `last_updated` (which is 95% of the runs) currently it's still skipped, since MistEO and later me, setup the quite simple Powershell script that checked for one line file changes that skipped the push to stop version.json from being modified on every run.

Now, the js file (and the ps1 as well) revert the changes in case no files are found to be modified, or the version.json has only `last_updated` modified. 

Don't they do the same exact thing anyway?
Unless there's also another one line change in which before it wasn't recognizable, now with the new implemented logic, it is.

AnnAngela:
@Constrat But previous pwsh script didn't prevent commit that only introduced one line changed for `last_updated` in `version.json` for specific server.

For example, f65210e changed `stages.json` for mainlandChina server, but it introduced unneed 1 line changes for other servers (`resource/global/YoStarEN/resource/version.json` etc.), which the js script will prevent it like 854c8af (no more `resource/global/YoStarEN/resource/version.json` if nothing changed in YoStarEN).

Another example is ae923d6, which changed activity of txwy only, but introduced unneed changes for all other server.

I check the new pwsh script by my eye only 😂, and looks good to me.

AnnAngela (dismissed):

AnnAngela:
TBH, I wanted to use pwsh at first, but I'm more used to nodejs so I used it 😂

tools/ResourceUpdateValidator/script.ps1
```diff
@@ -0,0 +1,108 @@
...
+
+# Improved logging for parsed diff
+Write-Output "Parsed diff result:"
+foreach ($server in $listPerServer.Keys) {
+    Write-Output "${server}: ["
+    foreach ($entry in $listPerServer[$server]) {
+        Write-Output "  [ '$($entry[0])', '$($entry[1])', '$($entry[2])' ],"
+    }
+    Write-Output "]`n"
+}
```
AnnAngela, line 46:
```suggestion
$listPerServer | Convertto-Json
```

Maybe more concise

AnnAngela:
> 15 seconds more

Seems be introduced by `actions/setup-node` and I think this action can be removed since the gha runner already has nodejs v20 (the `actions/setup-node` just my bad habit)

ABA2396 (dismissed):

Constrat:
@AnnAngela Thank you for your answers, it definitely makes more sense now.

Regarding the stages.json unfortunately I don't have any solution on this issue:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/commit/854c8af83f4ffafdf0abd5cda3016c931599eeaa introduced changes for the global (The new R6 collab) so technically the resource data should be updated for global only but since the stages.json is for all clients there's not really much we can do.

> Seems be introduced by actions/setup-node and I think this action can be removed since the gha runner already has nodejs v20 (the actions/setup-node just my bad habit)

You are right, maybe I was too fast to hate on the JS solution. I'll research a bit more to see what's the actual fastest :)

Constrat:
image
Strange, skipping the node setup step will make the run step take quite a lot longer, perhaps it does a small setup node as well?

AnnAngela:
@Constrat I checked raw log but it doesn't make sense to me, there is a 12s gap that I cannot understand, so let's use pwsh =.=

```
2024-09-06T08:05:36.6944808Z ##[group]Run node tools/ResourceUpdateValidator/index.mjs
2024-09-06T08:05:36.6945387Z [36;1mnode tools/ResourceUpdateValidator/index.mjs[0m
2024-09-06T08:05:36.6978026Z shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
2024-09-06T08:05:36.6978491Z ##[endgroup]
2024-09-06T08:05:48.3624246Z [2024/9/6 08:05:45] Check git status...
2024-09-06T08:05:48.5000167Z [2024/9/6 08:05:48] git status:
2024-09-06T08:05:48.5000934Z On branch perf/res-pwsh
2024-09-06T08:05:48.5001513Z Your branch is up to date with 'origin/perf/res-pwsh'.
2024-09-06T08:05:48.5001852Z 
2024-09-06T08:05:48.5001961Z Changes not staged for commit:
```

Constrat:
I really don't like the Github Actions log. Even the raw logs are pretty much useless.
I noticed something very similar when I was running the `prettier` step.
It used to take ages running `npx prettier`. Luckily, I found an action that already does the prettier step and it's 2 seconds as you can see.
With the `npx prettier` it used to take 10-15 seconds.
image

Constrat:
I think I see the problem. It's probably the `git pull` from the adding files step.

AnnAngela:
I would like to attribute this to poor performance of the Windows server or runner used by gha.

AnnAngela:
And if you have time, I think we can translate the code of the tools used by this workflow to work on linux server

Constrat:
Definitely, that's why I was working on converting the resource updater to a CMake build, but my knowledge is very limited on CMake, so I pretty much surrendered XD

AnnAngela:
And I don't understand why such tools need to be written in a compiled language, wouldn't python or nodejs be better?

Constrat:
I've "fixed" the compilation part by just caching the exe itself. 
ResourceUpdater is a very old tool and has always been in `cpp`. That way meojson was available con work on the various json files we are working on.

I know a bit of cpp so I was able to maintain it. (If we change it to py or js I wouldn't be able to do anything as I'm not really a fan of non-compiled languages)

Constrat:
Shame is that the ResourceUpdater was originally built as `.sln` when VS can also support native CMake, that would have made things a lot easier, allowing a seamless transitioning from Windows to Linux.

Constrat:
@AnnAngela The `--unshallow` option was the culprit.
Without it it takes 13 seconds less. I don't really understand the necessity of the unshallowing it. For the eventual rebase we just need the latest pull, not the entirety of the 
image

AnnAngela:
> I've "fixed" the compilation part by just caching the exe itself. ResourceUpdater is a very old tool and has always been in `cpp`. That way meojson was available con work on the various json files we are working on.
> 
> I know a bit of cpp so I was able to maintain it. (If we change it to py or js I wouldn't be able to do anything as I'm not really a fan of non-compiled languages)
> 
> Anyway I think it would be easier to just convert the build to a linux based one instead of having to completely rewrite the entire cpp 1300 lines to either python or js. I also think the performance would deteriorate since cpp itself is quite fast and currently the resource updater step is the slowest of them all.
> 
> The perfect solution would be to just convert the project to unix, that way the ubuntu gha runner _should_ be faster for all git checkout and git add git push operations.

Great, let me know if you need any help.

> @AnnAngela The `--unshallow` option was the culprit. Without it it takes 13 seconds less. I don't really understand the necessity of the unshallowing it. For the eventual rebase we just need the latest pull, not the entirety of the branch.

I agreed w/ you, I don't think we need all this history, I believe that `--depth 1` is just enough.

Constrat:
> I agreed w/ you, I don't think we need all this history, I believe that `--depth 1` is just enough.

Yep, just removing `--unshallow` is enough as the actions/checkout defaults to `fetch-depth: 1` in our case.

> Great, let me know if you need any help.

As you can see in https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/9943 I pretty much did all of the necessary work (I think) but the performance of the file is insanely slow + there's a stupid issue regarding the TilePos jsons.

I don't think I can do anything more as my CMake knowledge is zero. 
I just asked Github Copilot to do it and I just fixed all the errors cmake gave me, meaning there's 0 compiler optimizations and probably even some issues.

You said you don't know cpp, but maybe someone else from the MAA Team can have a look :)

PR #10690, opened 2024-09-27
fix: YostarKR switch OD navigation to template

HX3N:


Constrat (approved):

ABA2396:
可以参考一下 #10691 ，-Open 可以直接用现有的模板图了

Constrat:
So what modification do you suggest? Just adding the template without any other changes?

ABA2396:
之后简中会改成这样

``` json
"baseTask": "SS-Open",
"template": "StageSideStory.png",
"next": xxx
```

然后外服的 -Open 应该就不需要再有修改了，只需要修改 -OpenOcr 和对应的章节名应该就可以

ABA2396:
可以测试一下，在这个 pr 修改的 `-Open` 里添加 `"template": "StageSideStory.png"`，然后删掉新加的那张图看看能不能跑

Constrat:
So we will be able to remove all the previous xx-Open templates? (We never had a x2 rerun, but I guess for legacy and memories we are keeping them?)

ABA2396:
It should be possible. If the test is feasible, all previous navigation can be modified to this form.

ABA2396:
The probability of x2 is also very small. If really meet it, just use "current/last"

Constrat:
@ABA2396 Why is there a need for rewriting the name of template on globals?
Shouldn't baseTask do it already?

ABA2396:
> Any parameter that is not explicitly defined uses the value of the baseTask parameter for the corresponding task without a prefix, except for template which remains "taskName.png" if it is not explicitly defined.

The older version of the navigation only included "-Open." Whether it was template matching or OCR was determined by the specific task. Later, it was split into "-Open" and "-OpenOcr." "-OpenOpt" is just an overall entry point, and it should look like this:
``` json
"SS-OpenOpt": {
    "algorithm": "JustReturn",
    "next": ["SS-OpenOcr", "SS-Open"]
}
```
However, since "SS" needs to be replaced with the corresponding activity abbreviation, it doesn't really serve as a base task, as it will ultimately be rewritten anyway.

ABA2396:
The purpose of "-OpenOpt" is that during replication, whether entering from the bottom left corner as usual or from the middle when activities overlap, no existing content needs to be modified. When accessing this activity in the overseas-client, only the OCR text needs to be adjusted to provide support.

Constrat:
The only thing that I think could be an issue is the overlap of events. OD-Open doesn't exist in official because you guys had the rerun with nothing in the way, Yostar instead decided to throw at us the CC, hence the need to change and use the StageSidestory.

ABA2396:
Yes, that's why it's important to format the new navigation to https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues/10691 afterward, so that it's much easier to adapt afterward

PR #10733, opened 2024-10-04
ci: Resource Updater push perfs

Constrat:


Constrat:
image
Finally bypass also works on web and not only on mobile.
GitHub Support confirmed the deployment of the fix last night.

@MaaAssistantArknights/maacore-developers

PR #10734, opened 2024-10-04
docs: 明确文档中的 MacOS 称呼

Daydreamer114:


docs/zh-cn/manual/device/macos.md
```diff
@@ -9,9 +9,9 @@ icon: basil:apple-solid
 
 ### ✅ [PlayCover](https://playcover.io)（原生运行最流畅 🚀）
 
-试验性支持，遇到问题请多多提 issue，并在标题中提及 iOS。
+试验性支持，遇到问题请多多提 issue，并在标题中提及 `Mac` 。
```
Constrat, line 12: https://github.com/MaaAssistantArknights/issue-checker/blob/08d909e163099974c55f9dda40e45ac668bf91dc/.github/issue-checker.yml#L303-L304

This definitely clears up the idea of using MAA on iOS (which is the mobile OS) compared to MacOS which is the computer one, where MAA actually runs, but just as a heads up, the issue-checker seems a bit picky, especially on what capitalization is used. For example, this is not correct :P
image

Constrat (dismissed):

docs/zh-cn/manual/device/macos.md
Daydreamer114: then fix the issue checker. I forgot to update the en version for this doc.

Constrat: Whenver you have time :)

Rbqwow: our issue-checker's regex engine cannot directly specify to ignore case
so can only make it like `[Mm][Aa][Cc][Oo][Ss]`
holy sh*t

Constrat (approved):

Rbqwow:
we can use `(?i)` to ignore case now...?

AnnAngela:
JavaScript didn't support mode modifier like `(?i)`

Constrat:
:(

Jisu-Woniu:
> JavaScript didn't support mode modifier like `(?i)`

But `/ReGeX/i` is valid in JS.

AnnAngela:
> > JavaScript didn't support mode modifier like `(?i)`
> 
> But `/ReGeX/i` is valid in JS.

yes, use mode *flags* is allowed (e.g. `/foo/i`), but use mode *modifier* is not (e.g. `/(?i)foo/` or `/Re(?i)gex/`) in Node.JS

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Modifier#browser_compatibility

image

PR #10767, opened 2024-10-10
fix: YostarKR tweak Sami StageEnterBattleAgain specificRect

HX3N:
and tweak `CharsNameOcrReplace` ocr


HX3N:
GIF

Yostar changed the UI (only for Sami) :(

Alan-Charred (approved):
At first glance I thought we also need to adjust tasks in the form of `Sami@Roguelike@Stage...Enter`. Then I realised that this `EnterBattleAgain` task is the only unguarded one, which explains why the `specificRect` needs to be narrowed.

PR #10790, opened 2024-10-12
fix: EN Babel OCR ROI

swablueme:
This change modifies the OCR bounding box for the EN Babel Event

image


Constrat:
@HX3N This is what you mentioned before yes?

Constrat (approved):

swablueme:
thanks :)

HX3N:
> @HX3N This is what you mentioned before yes?

Yes, that's exactly right

PR #11007, opened 2024-11-01
fix: 基建换班宿舍任务期间重进宿舍后，重置过滤状态记录

status102:
- link #10968


status102:
cc @Constrat

Constrat:
Yes, I think this is much cleaner, instead of doing every single time for every dormitory.
Basing it on_run_fails() is much better.

PR #11080, opened 2024-11-05
feat: 基建新增“启用换班”选项以控制是否由MAA进行换班（默认启用）

EvATive7:


Constrat:
GUI is very wonky this way.
Morale bar is one more step under.
{AFC82C4A-E898-4B5A-9F52-148D1A846136}
{156BAA74-320B-4731-A349-AC56A7D53D50}

EvATive7:
Alright, I'll move the "Drone Usage" above the "Enable Shift"; By the way, the GUI of MAA has already become quite wonky.

Constrat:
> By the way, the GUI of MAA has already become quite wonky.

Yes this is indeed true, we now have quite a lot of options to be displayed in a 3 siderbars / and only one settings bar

EvATive7:
@Constrat OK

src/MaaCore/Task/Infrast/InfrastControlTask.cpp
```diff
@@ -4,6 +4,11 @@
 
 bool asst::InfrastControlTask::_run()
 {
+    if (!m_shift_enabled) {
+        Log.info("skip shift " + facility_name());
+        return true;
+    }
+
```
status102, line 11: 或许可以直接禁用对应任务

EvATive7: 其实一开始想的是每个房间都包含两个部分：房间单独业务+换班，比如会客室就是接线索+换班。
因此是直接跳过换班的，对应房间的单独业务由其剩下的代码决定。
不过确实很多房间是没有自己的单独业务的，如果直接禁用这些任务的话，会不会有点不太合逻辑？
这部分再讨论一下？禁用的版本我刚刚写了，讨论一下如果合适的话我再推上去。

status102:
有点怪，禁用换班也不应该让方舟自己一键换

Constrat:
The latest commit now tries to do two different things in the same PR. I don't suggest doing this as it becomes hard to maintain especially for global services.

Currently global doesn't have the autoshift directly in arknights meaning big changes in the MaaCore just for the Official are not ideal in the slightest. 
For this breaking changes I suggest separating PRs, this way we can add the necessary fixes / workarounds for the globals as well fixing the official.

EvATive7:
@Constrat 
Actually what I want to do is making MAA to support the latest autoshifting feature. If I want to implement this feature properly, the design of MAA infrastructure needs some reconstruction (including GUI, as you said before), but it is not easy, so I try to achieve the goal in a simpler way, but it is actually a bit weird. I plan to try to solve this problem using the method you mentioned above, but it may be a big change.  
I'm going to close this pr soon and start another pr to do this thing. Thank you for your suggestion.

PR #11303, opened 2024-12-06
feat: enhance MAA website animation for a more elegant appearance

chisato2233:
你好！我稍微更改了一下MAA官网的动画表现，集中在刚打开官网时的动画，之前没有太多的动画，现在看起来更加优雅。改动一般都是再各个组件上应用frame motion的blur淡入效果以及一些小的移动动画，希望你们能够喜欢！祝能够有一个美好的一天

### DIFF
--- a/website/apps/web/src/App.tsx
+++ b/website/apps/web/src/App.tsx
@@ -2,7 +2,7 @@ import { HomeHero } from './components/home/HomeHero'
 
 function App() {
   return (
-    <main className="h-full">
+    <main className="h-full max-h-screen overflow-hidden">
       <section className="h-screen min-h-[20rem] w-full relative">
         <HomeHero />
       </section>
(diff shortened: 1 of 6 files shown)


Constrat:
I'm trying the website locally. I don't really notice any changes?

chisato2233:
I didn't make many changes, so they might be subtle and easy to miss since they occur right at the start of the website loading. You'll notice a nice fade-in effect when you open or refresh the site—that's what I worked on. Comparing it to the live MAA website might make the difference more noticeable

Constrat:
Ok I see what you meant.

I tried it now and I see the change. But after trying it seems weird.

As you soon as you enter the website on `/docs/` the animation is weird as it goes away (to the top) and comes back (to the bottom).

Also the same effect happens on the live website when going from `/docs/zh-cn/` to `/docs/`

chisato2233:
My changes don't appear to affect the docs or their pages, so I'm unsure about what might have happened. Could the code have caused a bug in the docs? I ran the code on my computer and didn't notice anything unusual. I'd really appreciate it if you could provide more details!

chisato2233:
Oh, I think I understand what you're referring to—the scrollbar that briefly appears and then disappears, right? This occurs because initially, the webpage's height exceeds the browser's view, but shortly after, an animation adjusts it back to fit the browser's height. It's not ideal from an animation perspective. I'll work on fixing it and might submit  once it's resolved.

PR #11317, opened 2024-12-08
ci/nightly-schedule

Constrat:


ABA2396:
How about changing the value of limit as well? It feels like 30 iterations are completely unnecessary, and the packaging process takes a lot of time. I think 10 versions should be more than enough.

Constrat:
Why is the limit 30 anyway? It's not like that's the fetch checkout limit.

ABA2396:
It will generate OTA packages for 30 versions. If someone really hasn't updated for several months, they might not find an update package. However, the UI is designed to handle this by directly downloading the full package.

Constrat:
Default limit is now 10.
The backup value is also 10 in case weird stuff is inputted.

PR #11355, opened 2024-12-11
ci: disable optimize-templates on PR merge

Constrat:


Constrat:
Shitshow of commits, time to squash

PR #11408, opened 2024-12-21
feat: RA2 multiple crafting

Constrat:
@Alan-Charred I started farming RA as I was very behind and this is my result in just about 1 hour:
image
Pretty good right?

But then I thought. What are the most efficient crafting recipes? And so I started crafting just the net hunter launcher thing. But I was leaving the run with 300 wood never to be used.
And so I thought about just copying the logic like a retard to use both rocks and wood (for very advanced players you could even use more crafting with metal)

Obviously my implementation is super bad and just an idea. But have we ever considered adding more crafting recipes / option to craft more stuff?

### DIFF
--- a/src/MaaCore/Task/Reclamation/ReclamationCraftTaskPlugin.h
+++ b/src/MaaCore/Task/Reclamation/ReclamationCraftTaskPlugin.h
@@ -26,7 +26,7 @@ private:
         Hold = 1
     };
 
-    std::string m_tool_to_craft = "荧光棒";                // 要组装的支援道具
+    std::vector<std::string> tools_to_craft;                // 要组装的支援道具
     int m_num_craft_batches = 16;                          // 支援道具组装批次数, 每批组装 99 个
     IncrementMode m_increment_mode = IncrementMode::Click; // 点击加号按钮增加组装数量的方式
 };
--- a/src/MaaCore/Task/Reclamation/ReclamationCraftTaskPlugin.h
+++ b/src/MaaCore/Task/Reclamation/ReclamationCraftTaskPlugin.h
@@ -27,7 +27,7 @@ private:
     };
 
     std::vector<std::string> m_tools_to_craft = { "荧光棒" }; // 要组装的支援道具
-    int m_num_craft_batches = 16;                            // 支援道具组装批次数, 每批组装 99 个
-    IncrementMode m_increment_mode = IncrementMode::Click;   // 点击加号按钮增加组装数量的方式
+    int m_num_craft_batches = 16;                             // 支援道具组装批次数, 每批组装 99 个
+    IncrementMode m_increment_mode = IncrementMode::Click;    // 点击加号按钮增加组装数量的方式
 };
 } // namespace asst
--- a/src/MaaCore/Task/Reclamation/ReclamationCraftTaskPlugin.h
+++ b/src/MaaCore/Task/Reclamation/ReclamationCraftTaskPlugin.h
@@ -26,8 +26,8 @@ private:
         Hold = 1
     };
 
-    std::vector<std::string> tools_to_craft;               // 要组装的支援道具
-    int m_num_craft_batches = 16;                          // 支援道具组装批次数, 每批组装 99 个
-    IncrementMode m_increment_mode = IncrementMode::Click; // 点击加号按钮增加组装数量的方式
+    std::vector<std::string> m_tools_to_craft = { "荧光棒" }; // 要组装的支援道具
+    int m_num_craft_batches = 16;                            // 支援道具组装批次数, 每批组装 99 个
+    IncrementMode m_increment_mode = IncrementMode::Click;   // 点击加号按钮增加组装数量的方式
 };
 } // namespace asst
--- a/src/MaaCore/Task/Reclamation/ReclamationCraftTaskPlugin.h
+++ b/src/MaaCore/Task/Reclamation/ReclamationCraftTaskPlugin.h
@@ -26,7 +26,7 @@ private:
         Hold = 1
     };
 
-    std::vector<std::string> tools_to_craft;                // 要组装的支援道具
+    std::vector<std::string> tools_to_craft;               // 要组装的支援道具
     int m_num_craft_batches = 16;                          // 支援道具组装批次数, 每批组装 99 个
     IncrementMode m_increment_mode = IncrementMode::Click; // 点击加号按钮增加组装数量的方式
 };
--- a/docs/en-us/protocol/integration.md
+++ b/docs/en-us/protocol/integration.md
@@ -313,7 +313,10 @@ For more details about auto-copilot JSON, please refer to [Copilot Schema](./cop
                                 // 0 - Farm badges & construction pts (exiting the stage immediately)
                                 // 1 - Fire Within the Sand: Farm Crude Gold (forging Gold at headquarter after purchasing water)
                                 //     Tales Within the Sand: Automatically craft items and load to earn currency
-    "tool_to_craft": string,    // Automatically crafted items, optional, glow stick by default 
+    "tools_to_craft": [
+        string,                 // Automatically crafted items, optional, glow stick by default
+        ...
+    ] 
                                 // Suggested fill in the substring
     "increment_mode": int,      // Click type, optional. 0 by default
                                 // 0 - Rapid Click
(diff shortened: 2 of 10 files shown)


Constrat:
For a much cleaner and better implementation a foreach is required that loops for all the values of multiple `m_tool_to_craft`.
Then we need to think about a GUI way to allow at least one and optionally more user inputs for more crafts (we need to find a way to order them as the logic is just scroll down, no scroll up)

Alan-Charred:
> Obviously my implementation is super bad and just an idea. But have we ever considered adding more crafting recipes / option to craft more stuff?

Someone else mentioned this several months ago when this reclamation-bonus event occurs on zh_CN. There are some craft recipes not needing wood... and the initial idea is to craft them after glow sticks.
Unfortunately this improvement was put aside soon after the event...
If you want we can quickly pick it up.

> But I was leaving the run with 300 wood never to be used.

This can actually be controlled by the parameter num_craft_batches，which currently lacks a GUI-input.

> For a much cleaner and better implementation a foreach is required that loops for all the values of multiple m_tool_to_craft.

Exactly. And actually I was imagining an alternate list of tool names and num_craft_batches.
We can choose to order those items from top-down (either force the users to order them or we order them in para parsing phase). Or, we can choose to reset the list position before starting crafting a new item... just a little time waste.

And for gui input... the quickest way is to copy and modify the code for user-additional for copilot.

```
UserAdditional = UserAdditional.Replace("，", ",").Replace("；", ";").Trim();
            Regex regex = new Regex(@"(?<=;)(?<name>[^,;]+)(?:, *(?<skill>\d))?(?=;)", RegexOptions.Compiled);
            JArray userAdditional = new(regex.Matches(";" + UserAdditional + ";").ToList().Select(match => new JObject
            {
                ["name"] = match.Groups[1].Value.Trim(),
                ["skill"] = string.IsNullOrEmpty(match.Groups[2].Value) ? 0 : int.Parse(match.Groups[2].Value),
            }));
```
It parses an alternate sequence to json, which is pretty handy.

The most-frustrating part to me in this approach is, we need to consider compatibility with old input formats.
I guess we need a new param named... craft_plans? or something, in the form of list<pair<string, unsigned>>, and use the old params (tool_to_craft, num_craft_batches) only if craft_plans is not given.

Constrat:
I wasn't thinking that far ahead actually. I just though about a simple:
how many crafts do you want? 1 2 3 etc... USE AT YOUR OWN RISK
3 possible items. The user inputs them and needs to make sure they are already ordered.
Now the `m_tool_to_craft` is a vector and the foreach cycles through the necessary craftings (which are already ordered).
"Simple" as that. Instead of having the necessity of the copilot custom operator list.

Alan-Charred:
> I wasn't thinking that far ahead actually. I just though about a simple: how many crafts do you want? 1 2 3 etc... USE AT YOUR OWN RISK 3 possible items. The user inputs them and needs to make sure they are already ordered. Now the `m_tool_to_craft` is a vector and the foreach cycles through the necessary craftings (which are already ordered). "Simple" as that. Instead of having the necessity of the copilot custom operator list.
> 

That makes a lot of sense.

> Also, regarding the `m_num_craft_batches`. I checked the AsstProxy/AsstAppendReclamation and the user can select the value. I tried it around but I couldn't really find differences XD. I set it to one and it did 8 minutes of runs. I set it to 16 and I stopped it manually after 1 hour and 10 minutes :P

This argument should control the number of times we craft. So, if you have 300 woods unused with num_craft_batches=16, then these woods could be used up with a larger num_craft_batches...
Yes it does not save much time, but it should be used in this way... I will test it later, double-checking whether it still works as desired.

Alan-Charred:
> This argument should control the number of times we craft. So, if you have 300 woods unused with num_craft_batches=16, then these woods could be used up with a larger num_craft_batches...
> 
> Yes it does not save much time, but it should be used in this way... I will test it later, double-checking whether it still works as desired.

Just tested and the params works on my side.

If there are always plenty of ingredients left under a larger num_batches, please let me know...In that case we could meet with a bug.

Cheers,

Constrat:
Just to get an idea. If I put 1. Does that mean it will craft 99 times? 2 -> 99 + 99? etc?

Alan-Charred:
> Just to get an idea. If I put 1: does that mean it will craft 99 times? 2 -> 99 + 99? etc?

Exactly. And once MAA found you can only craft, let's say, 98 items instead of 99, then it knows the we are out of ingredients and should skip the "unused" batches.

(Let me quickly check the code)

```C++
if (craft_amount < 99) {
    // click the button again and make sure the craft amount has reaches its limit
    int old_craft_amount = craft_amount;
    ProcessTask(*this, { theme + "@RA@IncreaseCraftAmount" }).run();
    sleep(500);
    if (!calc_craft_amount(craft_amount) || craft_amount <= old_craft_amount) {
        insufficient_materials = true;
    }
    else {
        Log.info("Reclamation Craft", "| craft amount: ", craft_amount);
    }
}
```
Actually here it is a bit twisted...
After clicking the "+" button certain times / holding the button for a while, MAA checks the number on the screen. If it is lower than 99, MAA try click the button once again. If the number remains the same, then MAA turns on the flag `insufficient_materials`.

If the number changes, MAA knows it will continue to interact with the button to top it up to 99 before clicking the "craft" button.

Cheers,

Alan-Charred:
> > Just to get an idea. If I put 1: does that mean it will craft 99 times? 2 -> 99 + 99? etc?
> 
> Exactly. And once MAA found you can only craft, let's say, 98 items instead of 99, then it knows the we are out of ingredients and should skip the "unused" batches.
> 
> (Let me quickly check the code)
> 
> ```c++
> if (craft_amount < 99) {
>     // click the button again and make sure the craft amount has reaches its limit
>     int old_craft_amount = craft_amount;
>     ProcessTask(*this, { theme + "@RA@IncreaseCraftAmount" }).run();
>     sleep(500);
>     if (!calc_craft_amount(craft_amount) || craft_amount <= old_craft_amount) {
>         insufficient_materials = true;
>     }
>     else {
>         Log.info("Reclamation Craft", "| craft amount: ", craft_amount);
>     }
> }
> ```
> 
> Actually here it is a bit twisted... After clicking the "+" button certain times / holding the button for a while, MAA checks the number on the screen. If it is lower than 99, MAA try click the button once again. If the number remains the same, then MAA turns on the flag `insufficient_materials`.
> 
> If the number changes, MAA knows it will continue to interact with the button to top it up to 99 before clicking the "craft" button.
> 
> Cheers,

So one batch means 99 items and click "craft" button *once*.

Constrat:
@Alan-Charred I've checked again. The batch craft works as excpeted. I thought it was something else.

Constrat:
Btw, instead of using the copilot. This one is much simpler, (credit blacklist) and closer to what we want.
```cs
            var blackList = SettingsViewModel.MallTask.CreditBlackList.Split(';', '；')
                .Select(s => s.Trim());
```

Constrat:
@Alan-Charred Ok, I've used the blacklist MallSettings logic to use `;` as separator. I think I'm missing the case where the value is empty. But thinking about it it's basically impossible since the input can't be empty. It's hardcoded in the GUI section.

Constrat:
{1354EE02-E2F1-4B83-B574-D5394DE4A746}
Still. I don't understand why it does this. 16 Crafts set. It only does 2 (a single run) and leaves.

Alan-Charred:
> @Alan-Charred I've checked again. The batch craft works as excepted. I thought it was something else.
> 
> The only thing is. It leaves the gamemode after crafting. For example I put 2 as max craft per round.
> 
> It crafts 99 + 54 glow sticks. Skips -> skips -> Reset. Task completed

The task should only complete when MAA found that you have reached maximum points needed... which is not a good design for the event since people need to craft stuff anyway.

Well you can modify tasks.json to bypass this.. just delete the task ...ProsperityReachLimit from the next list of Tales@RA@PIS-FinishCalc.

Constrat:
> reached maximum points needed

OH. I didn't know we had the prosperity logic max logic XD. It's actually pretty neat if you ask me :)

Alan-Charred:
Give me some time and I will come back to you tomorrow. Cheers,

Constrat:
Take all the time you need. 
Christmas is close. Let's spend time with our families (I'll be online 24/7 probably XD)

Alan-Charred (approved):
Look all good to me.
I will add macGui adaptation later in another PR (in another repo actually since macGui is actually a submodule).

@wangl-cc will help with maa-cli (thanks a lot!)

Merry Christmas everyone~

Constrat:
Yeah I fucked up a bit the variables as it was just a WIP to see if everything was working, and it all worked first try so I forgot the cleanup XD

src/MaaCore/Task/Reclamation/ReclamationCraftTaskPlugin.h
```diff
@@ -26,8 +26,8 @@ class ReclamationCraftTaskPlugin : public AbstractReclamationTaskPlugin
         Hold = 1
     };
 
-    std::vector<std::string> tools_to_craft;               // 要组装的支援道具
```
Constrat: @Alan-Charred I didn't actually notice this lol.
If we initialize the vector like this. Nothing works :P

Constrat: I'll temp fix by not initializing. Do we actually need initialization?

PR #11446, opened 2024-12-28
ci: revert builds to ubuntu 22.04

Constrat:
All the remaining `ubuntu-latest` are mainly for logistic purposes.

I only changed the:
- `ci`
- `release-nightly-ota`
- `release-ota`

### DIFF
--- a/.github/workflows/release-nightly-ota.yml
+++ b/.github/workflows/release-nightly-ota.yml
@@ -247,7 +247,7 @@ jobs:
     strategy:
       matrix:
         target: [x64]
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Echo tag version
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/release-nightly-ota.yml
```diff
@@ -247,7 +247,7 @@ jobs:
     strategy:
       matrix:
         target: [x64]
-    runs-on: ubuntu-latest
```
wangl-cc, line 250: I believe we should leave this unchanged, as it's only for release, not for building.

.github/workflows/release-ota.yml
```diff
@@ -87,7 +87,7 @@ jobs:
 
   make-ota:
     needs: create-tag
-    runs-on: ubuntu-latest
```
wangl-cc, line 90: Same as the previous one.

wangl-cc (approved):
LGTM

PR #11476, opened 2025-01-02
fix: InitialDrop.png更新

BxFS:
现存的四个InitialDrop全是傀影那张InitialDrop的复制粘贴，更新了一下后三个肉鸽的InitialDrop

PR #11478, opened 2025-01-02
fix: InitialDrop.png更新

BxFS:
Continuation of #11476 per request of @Constrat


Constrat:
I just downloaded the files from your branch and pushed them directly, faster and without optimizing issues.

Thank you for your contributions to the MAA project!

PR #11480, opened 2025-01-02
chore: move `push tag` later in the workflow in case or errors

Constrat:
Had some troubles while testing a week ago, since the push tag step is done very early. 

In case of exceptions or weird errors in the workflow (build errors, actions breaking), the tag should be the last thing to push, since we would be creating "empty tags" with no referred nightly.

TL;DR
Push tag later = good.
I can't think of any bad side effect, since we use `steps.set_tag.outputs.tag' around the workflow to mention to TO-BE tag (and not `git describe --tags --match "v*"`)

### DIFF
--- a/.github/workflows/release-nightly-ota.yml
+++ b/.github/workflows/release-nightly-ota.yml
@@ -136,17 +136,6 @@ jobs:
 
           gh run watch ${{ github.run_id }}
 
-      - name: Push tag to main repo
-        id: push_main_tag
-        run: |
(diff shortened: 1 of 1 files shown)


Constrat:
Personal rant / comment. _Workflows get old quickly huh..._

I've been very much enjoying optimizing and looking for better solutions tho.

ABA2396:
我觉得没啥问题 
cc @AnnAngela

AnnAngela (approved):
LGTM

PR #11523, opened 2025-01-11
feat: skip task once and enable on end

Constrat:
Quite a lot of times I found myself manually changing the base infrastructure. This meant that I had to disable the base task and run all the other ones.

This lead to me forgetting about re enabling the base task and completely fucking the subsequent rotations (just like I did today :P )

This PR gives users the possibility to invert the right click functionality in the TaskQueueViewModel so that the task will be skipped on the first run and on Task:

MAA

### DIFF
--- a/src/MaaWpfGui/ViewModels/UI/TaskQueueViewModel.cs
+++ b/src/MaaWpfGui/ViewModels/UI/TaskQueueViewModel.cs
@@ -1057,7 +1057,7 @@ namespace MaaWpfGui.ViewModels.UI
             {
                 if (item.IsCheckedWithNull == null)
                 {
-                    item.IsChecked = false;
+                    item.IsChecked = GuiSettingsUserControlModel.Instance.InvertNullFunction;
                 }
             }
         }
@@ -1246,7 +1246,7 @@ namespace MaaWpfGui.ViewModels.UI
             int count = 0;
             foreach (var item in TaskItemViewModels)
             {
-                if (item.IsChecked == false)
+                if (item.IsChecked == false || (GuiSettingsUserControlModel.Instance.InvertNullFunction && item.IsCheckedWithNull == null))
                 {
                     continue;
                 }
@@ -1308,6 +1308,10 @@ namespace MaaWpfGui.ViewModels.UI
                 _runningState.SetIdle(true);
                 Instances.AsstProxy.AsstStop();
                 SetStopped();
+                if (GuiSettingsUserControlModel.Instance.InvertNullFunction)
+                {
+                    ResetTaskSelection();
+                }
                 return;
             }
 
@@ -1796,7 +1800,6 @@ namespace MaaWpfGui.ViewModels.UI
         private static bool AppendReclamation()
         {
             var toolToCraft = SettingsViewModel.ReclamationTask.ReclamationToolToCraft.Split(';', '；').Select(s => s.Trim());
-
 
             _ = int.TryParse(SettingsViewModel.ReclamationTask.ReclamationMode, out var mode);
 
--- a/src/MaaWpfGui/Configuration/GUI.cs
+++ b/src/MaaWpfGui/Configuration/GUI.cs
@@ -25,6 +25,8 @@ namespace MaaWpfGui.Configuration
 
         public bool UseNotify { get; set; } = true;
 
+        public bool InvertNullFunction { get; set; } = false;
+
         public string Localization { get; set; } = LocalizationHelper.DefaultLanguage;
 
         public bool MinimizeToTray { get; set; } = false;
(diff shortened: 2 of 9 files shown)


Constrat:
If this may be considered to dangerous for the generic user we can always just remove the .xaml part and only allow modification in the gui.new.json

status102:
Maybe we need a new icon style for not run once. The block fill style is used for execute once

Constrat:
The reason why I went with this approach is because WPF check by default only has 3 values
True, False and Undefined (null in our use-case).

My initial idea was to just extend the checkbox to a fourth state `SkipOnce`. That's fairly doable with an `enum` but after that I checked and in the xaml there's no way to "middle click" the checkbox to get another state since we already use right click for `OnlyOnce`.

status102:
Would you like some things like right click the gear button to run that single task?

Constrat:
> Would you like some things like right click the gear button to run that single task?

Oh I didn't think of that!

But how would that help? We still have only 3 states in the checkbox. Hence why I "inverted" the functioning with a bool

status102:
Just a new way to run task once.

Another scheme is to make taskQueue could like PostAction. But it's too heavy for such a multi items container.

status102:
I still have no idea with performing skip once. :(

Constrat:
The temporarily how about removing the GUI setting and just hardcode it in the gui.json? I quite need this feat and if it's "hidden" in the json it shouldn't lead to troubles since by default it's false.

ABA2396:
或许可以有一个类似完成后那样的 仅一次 功能，勾上之后就不保存之后的状态，结束后重置

Constrat:
But to reset means setting to false no? Which would kill off the purpose

ABA2396:
The "only once" function after completion feature restores the state to what it was before selecting the option. For example, if you had option A selected, then checked the "only once" option and later unchecked A, after the task is completed, A will revert back to being selected.

ABA2396:
But to be honest, I can't think of a good-looking way to add this checkbox to the main interface（

Constrat:
> The "only once" function after completion feature restores the state to what it was before selecting the option. For example, if you had option A selected, then checked the "only once" option and later unchecked A, after the task is completed, A will revert back to being selected.

This is not true. Looking at the logic it doesn't work like this and even trying it out. 
If I choose login task as selected and I right click, after running it will disable it since it follows this logic:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/bda924566f7f590ad6c8faf062c823121b9fbb02/src/MaaWpfGui/ViewModels/UI/TaskQueueViewModel.cs#L1054-L1063

The logic will set the state to false.

ABA2396:
image
I said can refer to this

src/MaaWpfGui/ViewModels/UI/TaskQueueViewModel.cs
```diff
@@ -1308,6 +1308,10 @@ public async void LinkStart()
                 _runningState.SetIdle(true);
                 Instances.AsstProxy.AsstStop();
                 SetStopped();
+                if (GuiSettingsUserControlModel.Instance.InvertNullFunction)
+                {
+                    ResetTaskSelection();
+                }
```
ABA2396, line 1314: The current implementation seems fine. If we were to modify it further, we'd need to add an enum, but it feels unnecessary. By the way, what is this section of code for?

Constrat:
But isn't that used for a differenthing:
the what to do after tasks are completed? I don't think it's the correct place, since there's no mention of tasks there. Anyway for now it's hardcoded in the config so if a user wants to change it they need to modify .json. Like for nightlies. Since checkbox is limited in states and we can't change since there would a shitton of work to do.

> we'd need to add an enum

It was the first thing I started going but then the checkbox wasn't compatible and I wanted to use middleclick to toggle "Skip Once" but in wpf it's not detected.

src/MaaWpfGui/ViewModels/UI/TaskQueueViewModel.cs
Constrat: We need to call `ResetTaskSelection` even if the count is zero since it's what I use to check the tasks:
```cpp
        public void ResetTaskSelection()
        {
            foreach (var item in TaskItemViewModels)
            {
                if (item.IsCheckedWithNull == null)
                {
                    item.IsChecked = GuiSettingsUserControlModel.Instance.InvertNullFunction;
                }
            }
        }
```

Without this, it would just skip the tasks (correct) but wouldn't select them. 
I nested it with the `GuiSettingsUserControlModel.Instance.InvertNullFunction` to make sure it only runs when we have the custom config set to true. Otherwise nothing changes from the usual.

Constrat:
@status102 Conflicts 😭

status102:
> @status102 Conflicts 😭 

sad to hear that :(

PR #11534, opened 2025-01-12
perf: 投资时检查源石锭是否不足

Daydreamer114:
用于避免强制retry 3次导致的时间浪费


status102:
ocr余额应该会好点，至少不用适配外服

Constrat:
As usual, I know it's annoying but when adding templates, send the full screenshot here (preferably with a border around the designed area). This way the adaptation is instant.

Daydreamer114:
i'll change it to ocr

Constrat:
Ok, but that's not the point. OCR is even worse since we just have text to look in another client that we probably don't even have access to.

So even for OCR a screenshot of where the text/template is located is very helpful.

Daydreamer114:
as status102 said, this task will try to ocr the ascii number, then no adaption is needed.

PR #11723, opened 2025-02-02
fix：招募测试函数修复

Roland125:
再出现资深就是真的有赛博女鬼了


Rbqwow (approved):

Constrat:
For future reference @Roland125 please use `:` instead of `：`

Rbqwow:
> For future reference @Roland125 please use `:` instead of `：`

he has already noticed it
just i squashed too fast :P

PR #11734, opened 2025-02-03
refactor: 重构公告

ABA2396:


ABA2396:
image
cc @Constrat 
Since we have switched to using ### as the delimiter, subsequent announcement content must start with a heading marked by three # symbols; otherwise, it may result in display errors.
But don't change it now, and then modify it together after the release of the version, otherwise there are too many announcements recently

Constrat:
Gotcha.
I also think we don't need the "temp announcement" display. We just need the long term one. But even then... We can just remove that as well and leave the permanent announcement there anyway.

Constrat:
The image itself is just for SKland yes? But I can just write something on top of it that says:
"Do not discuss MAA on official channels"

ABA2396:
Pretty much? I’m not entirely sure what communication channels are available on the foreign client, but the general idea should be close enough.

Constrat:
"Official" is enough. Biggest ones are basically just Discord and Twitter comments I guess, but the community one is Discord.

Constrat:
Lazy, did not write the logic

Constrat:
Also, I personally feel like `ALL~ the Announcements` is a bit childish.

src/MaaWpfGui/ViewModels/UI/AnnouncementViewModel.cs
```diff
@@ -26,6 +32,91 @@ namespace MaaWpfGui.ViewModels.UI
...
+                    return new AnnouncementSection
+                    {
+                        Title = lines.FirstOrDefault(),
+                        Content = string.Join("\n", lines.Skip(1)).TrimEnd('-'),
+                    };
+                }).ToList();
+
+            sections.Insert(0, new()
+            {
+                Title = "ALL~ the Announcements",
```
Constrat, line 87:
```suggestion
                Title = "All the Announcements",
```
Even though the ideal solution would be to have this **bold**

Constrat (approved):

Constrat:
I changed it already. I'm sorry but my OCD was acting. It was so bad having incosistencies.
Should we actually consider creating a prod for announcements as well?

Alan-Charred:
> Should we actually consider creating a prod for announcements as well?
> 
> 
> 
> Or perhaps a more complicated check on our end to decide when to publish a "new" announcement or not even though changes were made. (Users don't read announcements anyway)

Could you please elaborate this a further more? Are you suggesting creating a workflow for announcement？ (apologise I'm not so familiar with this "prod", I assume it is a workflow）

Constrat:
> Could you please elaborate this a further more? Are you suggesting creating a workflow for announcement？ (apologise I'm not so familiar with this "prod", I assume it is a workflow）

Nothing of the sort, forget about it. After thinking this through tonight it was my mistake preemptively pushing the announcement fix for EN.

Constrat:
@ABA2396 With the latest commit I'm noticing some inconsistencies: 
{7E0940E3-57F2-4F40-A766-DE1D2115D8E4}

{E6935C78-4305-4152-AC43-6B3D0F69A8F9}

Some have the `----` some don't.

ABA2396:
I looked at it because some wrote `---` and some wrote `----`.

ABA2396:
Wait a minute and see how to change it better.

SherkeyXD:
怎么pre commit又炸了

PR #11867, opened 2025-02-11
feat: bypass update resources in formatting cases

Constrat:
Wait before merge.

- [x] Basic functionality

Optimizations: 
- [x] Before data (git show HEAD and py tasksort)
- [x] Post data (prettier together with .exe artifacts)

HALF COMPLETED:
- [x] Tests with dummy commits

EDIT: Thinking about it I could completely remove the current version.json checker, to follow the new implemented logic that follows the tasks.json (If only json is modified, should version.json be modified?)

So my questions are as follows:

1. `tasks.json` and `version.json` are connected to eachother? tasks.json don't appear in MaaResource, so I don't think they should, but currently they do. If this PR gets merged as-is there will be inconsistencies between the main repo `version.json` and the MaaResource `version.json`, which can potentially lead to the mirrors being outdated compared to the nighlies.
2. When is update-resources supposed to run? Currently it runs on every commit (and is wrong) as there are some commits that are just formatting or sorting. Those commits DON'T / CAN'T need to be removed as they "fix" various issues implemented by us users when pushing to dev (or when precommit dies and doesn't fix a PR itself)

Depending on the answers I might need more time to fix / tweak / perf subsequent things related to the ResourceUpdater branch.

As an example, if the version.json logic needs to be changed I will modify validator.ps1 from scratch.
cc @MistEO @ABA2396 @AnnAngela


Constrat:
Runs:
- https://github.com/MaaAssistantArknights/MaaAssistantArknights/actions/runs/13270392807

Constrat:
Optimized, but it's still not enough. Last check is needed to bypass everything in case there are real files (for example in this [run](https://github.com/MaaAssistantArknights/MaaAssistantArknights/actions/runs/13274933901/job/37062565358), stages.json was modified but since tasks.json was empty, (no modifications at all) it still returned false. This case needs to be true, even if tasks.json has not useful changes. By bypass I mean to nest all the new validator.ps1 code in a `if` with perhaps a negative condition that checks if something other than tasks.json (and version.json) have been modified, if yes, then it will update anyway, if not, it will check if tasks are worth it.

AnnAngela:
I thought I had the software's update process figured out, but turns out I was completely wrong.

@ABA2396 学姐救我

MistEO:
The only purpose of `version.json` is to know the versions of [these files](https://github.com/MaaAssistantArknights/MaaResource/tree/main/resource) (note that this is different from the files in [main repo](https://github.com/MaaAssistantArknights/MaaAssistantArknights/tree/dev/resource), which only contains a few files) and get updates from out server, so we should update `version.json` only when these files are updated. For global, it does not include tasks.json (of course, it does not include tasks.json for cn either)

Constrat:
Just a style: format resources makes sense but how do you separate it from the current chore: Auto update resources? Running the .exe will show differences as it rewrites the jsons

Constrat:
Uhm difference between local tests and runner.

ABA2396:
突然想到一个，资源更新是不是可以带上tasks.json里干员名和关卡名翻译的内容？不然的话就算资源更新了还是不认识新干员，因为他们的名字并没有被成功翻译

Constrat:
ResourceUpdater already replaces the necessary OCR for the overseas.

Constrat:
I see, but regexing is needed

Constrat:
Ok I'm going crazy, why is the runner being a little bitch.
Differences detected in resource/global/txwy/resource/tasks.json, RUN UPDATE RESOURCES.
Differences detected in resource/global/YoStarEN/resource/tasks.json, RUN UPDATE RESOURCES.
Differences detected in resource/global/YoStarJP/resource/tasks.json, RUN UPDATE RESOURCES.
Differences detected in resource/global/YoStarKR/resource/tasks.json, RUN UPDATE RESOURCES.

vs local:
No substantial changes in resource/global/txwy/resource/tasks.json.
No substantial changes in resource/global/YoStarEN/resource/tasks.json.
No substantial changes in resource/global/YoStarJP/resource/tasks.json.
No substantial changes in resource/global/YoStarKR/resource/tasks.json.

Constrat:
I am stupid, actually I'm retarded. Of course it finds differences lmao, there's no .prettierrc

Constrat:
Laudatum et Christo
If works

PR #11872, opened 2025-02-12
fix: 月度小队模式不再试图提前离开肉鸽

BxFS:
回避现有流程先考虑是否退出再考虑其他task（比如StageEncounterSpecialClose）的问题，另外从萨米开始月度就不再是抵达四层即可结束，不如一直打下去


Daydreamer114 (approved):

Manicsteiner:
外服好像没有mode6和mode7

BxFS:
> 外服好像没有mode6和mode7

@Constrat Not very sure about how this works but I do remember you telling me about missing translation regarding mode 6 and mode 7 yesterday. Is this to be fixed on somewhere else, like for example i18n site?

Constrat:
@BxFS Technically yes, but webplate can be slow and unpredictable so the usual scenario is when new feats or strings are added, they are added for all i18n xamls (even in chinese for other languages, but at least they are there, so that when we check we see the missing translation).

Even if the stuff is not on the global servers translations can be done before (Except for specific names like Roguelike Squad names, pre translations are useful because otherwise the client will have chinese text on non translated pages and we wouldn't know about it.

PR #11875, opened 2025-02-12
feat: script to update version.json

Constrat:
@MistEO Ok, so shifting all logic to independent version.json modifications.

version.json will get modified when any file is modified except the tasks.json (follows [MaaResource files](https://github.com/MaaAssistantArknights/MaaResource))

~~Now I just need to revert and fix the logic to when we were previously checking for files differences and pngs, to correctly run the optimizations png step and then the commit and push.~~ DONE

Since I didnt completely understand the logic of when update-resources is run I will keep the bool that skips it, so for now the bool is directly connected to when version.json gets modified. If it gets modified, it will also run update-resources.

@AnnAngela After merge, do you think we delete validator.ps1 and mjs? Or do we keep them just in case?

### DIFF
--- a/.github/workflows/res-update-game.yml
+++ b/.github/workflows/res-update-game.yml
@@ -303,20 +303,17 @@ jobs:
           args: -w ${{ steps.task_sorting.outputs.gitdiff }}
 
       - name: Update version.json date if necessary
+        id: update_version
         run: pwsh tools/ResourceUpdater/version.ps1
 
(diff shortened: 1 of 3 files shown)


Constrat:
It works!

MistEO (approved):

AnnAngela (approved):
I think we can safely delete validator.*, no need any more(x

PR #11921, opened 2025-02-14
feat: YoStarEN Sarkaz roguelike

Constrat:
cc @Daydreamer114 if you want to have a look at it


Daydreamer114:
i'm working on these

resource/global/YoStarEN/resource/template/Screenshot_20250214-230530.png
Daydreamer114, line 1: 1

Daydreamer114 (changes requested):

resource/global/YoStarEN/resource/template/Screenshot_20250214-230530.png
Constrat: Oh no, I directly pasted the wrong thing (Basically the untouched screenshot lol)

Constrat:
@Daydreamer114 what about all the ocr tasks for the Thought thing 
(I remember for the Sami Foldartals it was quite a pain in the ass)

Constrat:
Idea? Whatever it's called. The first translation was `Idea` but it seems called `thought`

Constrat:
LMAO

Daydreamer114:
those icons are clear enough, I would consider using matchtemplate if we want to write further functions.

Constrat:
For foldartals we used OCR because they were stored in the roguelike_topic gamedata and so easily accessible via the resource updater

Daydreamer114:
this will create a lot of work for global adaptation. besides, since they really care about MaaRelease and MaaResource, I am no longer willing to use them for pushing fixes.

Daydreamer114:
what is this
image
image
image
image

Daydreamer114:
why the fk full match is true

Constrat:
Great question

Constrat:
@Daydreamer114 I've been trying to blame who added the full_match. I can't find it. I agree that it makes absolutely no sense in leaving it there, especially when the OCR gets a lot of garbage there.

Daydreamer114:
so if we disable "full match", "Something else" and "Something" can all match "Something". Is this acceptable? Or simply add "$" for "Something<?$"

Daydreamer114 (approved):
failed at Emergency Lecture, classic
LGTM

Daydreamer114:
btw we usually call Sarkaz IS4. CEOBE'S FUNGI MIST is IS0 :)

Daydreamer114:
merge? XP

Daydreamer114:
@MaaAssistantArknights/global-support-team @MaaAssistantArknights/maacore-collaborators 

我提议将不同肉鸽主题的 StageEncounterOcr 进行拆分。我们已经遇到了JP服有两个 ‘相遇’ ，我在重新检查 EN服时也遇到了相似的事件，以及可能导致重复识别的正则表达式。可以在 json 中写为 Sarkaz@Roguelike@StageEncounterOcr ，在 cpp 中直接添加 'theme + '@' + Roguelike@StageEncounterOcr'。

I propose to split the StageEncounterOcr for different IS themes. We have already found two 'encounters' on the JP server, and I have come across similar events when rechecking the EN server, as well as regular expressions that might lead to duplicate recognition. It can be written in json as Sarkaz@Roguelike@StageEncounterOcr, and directly added in cpp as 'theme + '@' + Roguelike@StageEncounterOcr'.

image
image
image
image

Constrat:
> btw we usually call Sarkaz IS4. CEOBE'S FUNGI MIST is IS0 :)

Oh no, here is different. IS2 phantom 3 mizuki 4 Sami 5 sarkaz

Constrat:
ok let's merge. In case issues arise. We recvert the full match

Daydreamer114:
> Oh no, here is different. IS2 phantom 3 mizuki 4 Sami 5 sarkaz

image
actually hypergryph themselves call it 4th roguelike

Constrat:
Interesting, anyway considering the encounter splitting, I need to change the ResourceUpdater logic

Daydreamer114:
then i'll create the pr

PR #11926, opened 2025-02-15
feat: remove spaces for all ocrReplaces for KR + various optimizations

Constrat:
@HX3N So, I've made some optimizations as well. Have you ever used the ResourceUpdater?

I would like you to test if what is done is what you want / expect it to.


Constrat:
Unfortunately since you've already done some work because if my mistake the "old" stuff can't be removed automatically with the new unspaced strings because it already detects them as existing.

HX3N:
image
~There were still some spaces remaining~

After deleting some items from the tasks and testing it, the process of removing the spaces worked well.
Thank you for doing this work 👍 👍

HX3N (approved):

Constrat:
Yes, as I told you, already present items will not get touched as the key already exists.

The two spaces that remain are another unicode character that I can't seem to be able to remove

PR #11956, opened 2025-02-17
feat: YostarEN I.S. modes

Constrat:


Daydreamer114 (approved):
nothing reviewable, just ocr replace )

Constrat:
Just in case I made some stupid mistakes with the base tasks.

Constrat:
Let's merge then!

Daydreamer114:
actually what I remembered is that if the KEY in cn is exist and the VALUE is the same, then that KEY can be omitted
but may not very appropriate for templates

Constrat:
That's why I'm keeping it, in some tasks I'm rewriting the ROI, su just to make sure, I reapply all the texts

PR #11960, opened 2025-02-17
fix: YostarKR updated the StartUpConnectingFlag template

HX3N:
``` python
[2025-02-17 18:32:10.404][ERR][Px70a0][Txa248] templ size is too large StartUpConnectingFlag.png image size: 450 85 templ size: 525 85
```

Yostar changed it :|


Constrat:
Unsurprising

PR #11962, opened 2025-02-17
Release v5.13.2

Constrat:
Sarkaz I.S. for Global


Daydreamer114:
I would prefer a beta version for less effect, gives us time to fix issues and make a final release on stable version )
now with mirrirc, each release may cost misteo at least 40 bucks

Constrat:
Whatever you think is best. I would like to release something non nightly asap since users can't read about sarkaz being in development

Daydreamer114:
emmm they said their server was broken……?

MistEO:
ota mirrors are ready

https://ota.maa.plus/MaaAssistantArknights/api/version/beta.json

MistEO (approved):

Constrat:
I still don't like the fact that ci.yml also prepares the release, but I'm also lazy to change it to another worjflow and call it from there.

Issue #11988, opened 2025-02-20
Currently non implemented/bugged feature for IS sarkaz EN: Monthly Squad


dragonheart107:
### Before raising the issue...

- [x] I understand that Issues are for feedback and problem solving,  
not trolling the comments section, and will provide as much information as possible to help solve the problem.

- [x] I filled in a short, clear title so that developers could quickly identify the general problem when going through the issue list.  
Instead of "Some suggestions", "Stuck", etc.

- [x] I am using the latest version of the current update version and have reviewed both the [Unreleased changes](https://github.com/MaaAssistantArknights/MaaAssistantArknights/commits/dev/) and the [Pull requests](https://github.com/MaaAssistantArknights/MaaAssistantArknights/pulls),  
there is no mention of the bug being fixed.

- [x] I have reviewed the [FAQs](https://maa.plus/docs/en-us/manual/faq.html), [Announcement](/MaaAssistantArknights/MaaAssistantArknights/issues/7732), [Open Issue](https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues), [Closed Issue](https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues?q=is%3Aissue%20state%3Aclosed) to ensure that my issue has not been mentioned.

### Description

Issue 1:
when both options ticked
![Image](https://github.com/user-attachments/assets/a7cbed87-0b1d-4da7-99b6-f6f184184f34)

MAA behaves like recording below, it gets stuck on the squad communication page and the log interface of MAA itself shows that exploration has been started

Issue 2: if only first option is ticked, it correctly starts a monthly squad run, however MAA gets stuck on the nymph page

### Version

Resource Version: A Kazdelian Rescue
Resource Time: 2025/02/20 10:06:52
UI Version: v5.14.0-beta.1
Core Version: v5.14.0-beta.1
Build Time: 2025/02/18 11:21:15

### Log and config files

[asst.bak.zip](https://github.com/user-attachments/files/18886670/asst.bak.zip)
[asst.log](https://github.com/user-attachments/files/18886591/asst.log) around 2025-02-20 13:17:57.808 for issue 1 and around 2025-02-20 13:19:50.527 for issue 2
[gui.log](https://github.com/user-attachments/files/18886592/gui.log)

[gui.json](https://github.com/user-attachments/files/18886686/gui.json)

### Configuration information

[...]

PR #11991, opened 2025-02-20
fix: add an missing @ for MonthlySquadCommsBackTwice

BxFS:
fix the fixable part of #11988

somehow CN can take it while EN can't, no idea why


BxFS:
@Constrat CN tested, would this work on EN

PR #12021, opened 2025-02-23
fix: fast forward text in encounter

Constrat:
@dragonheart107 You can use this if you want.
Make delay changes as you wish if they are working correctly

### DIFF
--- a/resource/tasks.json
+++ b/resource/tasks.json
@@ -7677,7 +7677,7 @@
     "Roguelike@StageSafeHouseSkipText": {
         "template": "empty.png",
         "action": "ClickRect",
-        "preDelay": 500,
+        "preDelay": 250,
         "specificRect": [0, 405, 725, 225],
         "next": ["Roguelike@StageSafeHouseSkipTextRewards"]
     },
--- a/resource/tasks.json
+++ b/resource/tasks.json
@@ -7684,7 +7684,6 @@
     "Roguelike@StageSafeHouseSkipTextRewards": {
         "template": "empty.png",
         "action": "ClickRect",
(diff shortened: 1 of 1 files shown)


resource/tasks.json
```diff
@@ -7671,10 +7671,23 @@
...
+        "template": "empty.png",
+        "action": "ClickRect",
+        "preDelay": 500,
+        "specificRect": [0, 405, 725, 225],
+        "next": ["Roguelike@StageSafeHouseSkipTextRewards"]
+    },
+    "Roguelike@StageSafeHouseSkipTextRewards": {
+        "template": "empty.png",
+        "action": "ClickRect",
+        "preDelay": 500,
```
Constrat, line 7687:
```suggestion
```

resource/tasks.json
```diff
@@ -7671,10 +7671,22 @@
...
             "Roguelike@StageSafeHouseEnter",
-            "Roguelike@StageEncounterOptions#next",
-            "Roguelike@NextLevel"
+            "Roguelike@StageSafeHouseSkipText"
         ]
     },
+    "Roguelike@StageSafeHouseSkipText": {
+        "template": "empty.png",
+        "action": "ClickRect",
+        "preDelay": 500,
```
Constrat, line 7680:
```suggestion
        "preDelay": 250,
```

status102:
what's this, why the template is empty

Constrat:
If you think it works, change it in dev directly.

status102:
> If you think it works, change it in dev directly.

thanks x) I may test it after beta is published ( so when would we merge it

PR #12076, opened 2025-03-03
ci: nightly版本打包tag获取

status102:
ci 低手


Constrat:
Weird, checkout was doing that correctly and `v4` changed nothing regarding that (even weirder), but sure we can merge and hope it works :)

Constrat:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/actions/runs/13625413057/job/38081673923
Works

PR #12087, opened 2025-03-05
fix: burden/toil tasks threshold 0.9 -> 0.8

BxFS:
Should fix the issue mentioned in discord about EN stuck at mental toil screen and not selecting loadbearing opers.
Untested, need testers.
Unknown if same problem exist for other global clients.


Constrat:
Of you are reducing CN as well. We can just:

resource/global/YoStarEN/resource/tasks.json
```diff
@@ -2918,7 +2918,8 @@
         "text": ["HYPOTHESES", "COMPILATION"]
     },
     "Sarkaz@Roguelike@ChooseBurdenOperatorAtStart": {
-        "roi": [589, 584, 231, 121]
+        "roi": [589, 584, 231, 121],
+        "templThreshold": 0.8
```
Constrat, line 2922:
```suggestion
```

resource/tasks.json
```diff
@@ -10595,7 +10595,7 @@
     },
     "Sarkaz@Roguelike@DecreaseBurdenRetardation": {
         "roi": [561, 607, 153, 113],
-        "templThreshold": 0.9,
+        "templThreshold": 0.8,
```
Constrat, line 10598:
```suggestion
```

Constrat (changes requested):

BxFS:
> Of you are reducing CN as well. We can just removed everything
> 
> The default is 0.8 anyway. Are you sure it's correct modifying CN as well?

That's two different tasks btw, and the EN only task have different template than CN so that's probably CN irrelevant; the task modified in CN task.json had same template and I'd assume CN would encounter same problem.

resource/global/YoStarEN/resource/tasks.json
BxFS: The corresponding task in CN has 0.9, defaulting in EN would probably be 0.9 as well

Constrat: You're right. I reviewed with mobile only.

Constrat:
Since it's just thresholds... We merge!

PR #12094, opened 2025-03-07
feat: SideStory「挽歌燃烧殆尽」导航

SherkeyXD:
roi要等开服才能测，先挂着


resource/template/infrast/Bskill_man_spd&trade.png
Constrat, line 1: Interesting change

PR #12127, opened 2025-03-10
chore: YostarKR DT stage roi and text

HX3N:


Constrat:
Oh no.

Constrat:
Just to get an idea. Is this similar to the Siracusano event? Where there's more navigation required?

Daydreamer114 (approved):

HX3N:
> Just to get an idea. Is this similar to the Siracusano event? Where there's more navigation required?

I’m not sure how Siracusano was, but you just need to add the stage name.

PR #12132, opened 2025-03-10
feat: 更新 243 极限效率一天四换排班表（20250310 修订）

bodayw:
根据[公孙长乐 2025-03-10 新视频](https://www.bilibili.com/video/BV1QARvYFEWi/?t=8)制作。

制造站微调，经验书产量有略微提高（食铁兽被优化掉了）。

Minor modifications for the factories. Battle records production slightly increased (FEater is no longer needed).


Constrat:
TL;DR: Thorns Alter instead of  FEater?

bodayw:
> TL;DR: Thorns Alter instead of FEater?

Yeah, basically by switching around a few operators in the factories, allowing Thorns Alter to work for eight hours instead of just four, and we then let Ulpianus to work in the battle record factory instead of FEater. Turned out we could have achieved slightly better production this way a couple of months ago haha.

Constrat:
@bodayw Sorry for the necro on this PR. But since it's time for EN to use Thorns Alter I've been using the "new" older CN base, after a bit fo tweaks like Ascalon instead of Amiya (just for when I will use the training room). I've noticed a bit of weird "sort" and "skips". For example Whisperain from the 4th rotation to the 1st remains in the office, but there's no skip. 

The reception room sorts Ines first. Is that to make sure shes does not reset her % counter? Changing her from 1 to 2 will make lose the % reception counter?

Again, about sorting. Waai Fu, Aroma and Gravel. Why is there a need for sorting there?

I've noticed in the 3rd dorm (Where Virtuosa, Czerny and Ines station) instead of placing Vigil in the 4th rotatiowe use the autofill, and only in the first rotation we finally place Vigil... Wouldn't it be best to already place vigil in the 4th rotation and then in the first rotation swpa Muelsyse for Bryophita?

Speaking of Bryophita, something similar happens during the 1st and 2nd rotation. IN the 2nd rotation we add Ch'en to the 3rd dorm, and leave the autofill, when we could just leave bryophita in that spot and just replace Vigil with Ch'en.

If you can help me clear these doubts I'd love it since arknights base is not that straightforward... Perhaps you know some secrets that I don't.

This is just a side comment since not a lot would change, but the order which the operators are placed in the dorms doesn't seem that optimal? For example, the 1st dorm will not have operators with the lowest morale, leading to MAA searching for those ops instead of placing "the first ones"...
This is just me min maxing, but in the long run this adds runtime.

Just to be clear. The work you do to translate the video to the json is amazing and I've been using the json (with slight tweaks) for the past year. I'm just commenting when the solution would be just to make my own base json like I used to two years ago.

bodayw:
@Constrat 

> Whisperain from the 4th rotation to the 1st remains in the office, but there's no skip.

I purposefully made it that none of the rooms in the 1st rotation will be skipped, just to make sure everything's in place when somebody uses the schedule for the first time.

> Is that to make sure shes does not reset her % counter? Changing her from 1 to 2 will make lose the % reception counter?

Precisely.

> Waai Fu, Aroma and Gravel. Why is there a need for sorting there?

Aroma's second base skill is also a "warming-up" one (+2% per hour, up to +20%).

> in the 3rd dorm (Where Virtuosa, Czerny and Ines station)

(guess you meant Iris instead of Ines)

> Wouldn't it be best to already place vigil in the 4th rotation and then in the first rotation swpa Muelsyse for Bryophita?

I guess it's because Muelsyse needs to go to work by the first rotation?

...and regarding to Vigil and Bryophyta, I can't recall why I did all those lol. This plus...

> the order which the operators are placed in the dorms doesn't seem that optimal? For example, the 1st dorm will not have operators with the lowest morale, leading to MAA searching for those ops

Basically for all the rooms except the dorms, I tried to make them exactly as in 公孙长乐's video, but for the dorms, I tend to put operators belong to some kind of squads together in the same dorm room, e.g. in the first rotation I put the five Abyssal Hunters in the same room, and Siege, Delphine, Morgan the three Victorians together, or sometimes it's just to let say, Purestream, Weedy, Eunectes who always work together stay together...I figured the operators might like it this way (if they were real lol). But in the end it's kinda random given there are gonna be some operators left alone and I had to put them wherever there's a free slot.

And it is the the way it is could also due to how it was in the previous version, because I typically work it out for the new version base on the previous one, instead of doing it from scratch.

So yeah, it might not be optimal in terms of MAA runtime. I'll prioritize this (over my preferred "squads") in future versions.

P.S. Feel free to nitpick!

Constrat:
Thank you very much for the answers. So regarding the dorms, I don't mind the squad composition. Since for example abyssals have buffs in dorms and Victoria as well. But mainly prioritize in the first dorms lower morale squads, that way you swipe less with MAA.

Constrat:
Also in the cases where you think autofill is necessary. Try to look if the previous operators in the dormitory can be left inside. (Like Bryophita for example)

PR #12168, opened 2025-03-16
feat: SSS#6 for EN

dragonheart107:


Daydreamer114:
@Saratoga-Official 这里全改成英文是不是不太好

Saratoga-Official:
复制一份吧，EN服的这么快就开了？

dragonheart107:
only the first few got translated and it shows as this anyways in EN:

grafik
 which is sufficent for 90% since we EN can't ready anyways

PR #12193, opened 2025-03-21
fix: YostarKR lower StartToVisit templThreshold

HX3N:
log
``` python
[2025-03-16 15:57:49.944][TRC][Px6824][Tx470c] match_templ | StartToVisit.png score: 0.732708 rect: [ 1057, 148, 59, 34 ] roi: [ 952, 105, 267, 601 ]
```

### DIFF
--- a/resource/global/YoStarKR/resource/tasks.json
+++ b/resource/global/YoStarKR/resource/tasks.json
@@ -1073,6 +1073,9 @@
     "NoFriends": {
         "text": ["등록된", "없습니다"]
     },
+    "StartToVisit": {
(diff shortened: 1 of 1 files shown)


Constrat:
For now, let's merge, but perhaps retaking the template might help?

HX3N:
> For now, let's merge, but perhaps retaking the template might help?

I get good scores, but I modified it like this because another user said it wasn't being recognized. I think it might be due to resolution or emulator differences.

Even if we retake the template, I feel like the result would be the same—unless the other users change their resolution or emulator. :/

PR #12244, opened 2025-04-04
feat: YostarJP GO navigation

Manicsteiner:
note: 本次更新 理智上限提升，助战系统更新，材料合成更新，基建更新 - from nga


Constrat:
Can you remember if we have to do anything zero day to fix the infrastructure? Like uncomments tasks etc etc

HX3N:
> Can you remember if we have to do anything zero day to fix the infrastructure? Like uncomments tasks etc etc

Isn't the only thing related to the infrastructure the Clue Vacancy image? 2a93ea01d2e4a80d4fa8cc50cb130e60eec87c73

Constrat:
New templates

PR #12269, opened 2025-04-07
fix: use kengxxiao for some files.

Constrat:


MistEO:
LGTM

But this is a temporary strategy, we still need yuanyan's resources in the end, kengxxiao's lack of map tile data

Constrat:
> LGTM
> 
> But this is a temporary strategy, we still need yuanyan's resources in the end, kengxxiao's lack of map tile data

Yes. We still use yuanyuan fir templates and levels.json (outdated) but at least the new jsons are are added. This is quite bothersome, as it also breaks all other clients. Perhaps we can add dummy templates temporarily to fix the smoke testing since new operators will be released today for global

MistEO:
Yuanyan said he will update his repository tonight (Beijing time). I guess he may update it within 8 hours. Can this catch up with the update of globals?

Constrat:
JP and KR will have the update in 4 minutes (or an hour, we have daylight saving time now so I'm not sure) but EN should be good at 19:00 my time in 9 hours

PR #12274, opened 2025-04-08
feat: overseas (JP & KR) data

Constrat:
@MistEO This also adds the battle_data for the new CN operators (Mon3tr, Alanna etc)

Running it locally, it also updated the stages.json with the new story line but I will not add them this this PR.

Technically the only issue is the infrast.json updates as they add json objects that don't exist as templates hence the smoke testing.
Infrast:
{F2657CC4-F1C6-4B55-BEEC-C4BFAF22E14A}
Stages:
{8B307876-B993-40CC-96B5-FF65C5D975DB}


Constrat:
I think we do need to update the main.cpp for the Resource Updater anyway since HG will probably not revert the new json object in the gamedata.

PR #12276, opened 2025-04-08
chore: YostarJP clue and GUI updates

Manicsteiner:
still different fonts


Constrat:
Different fonts?
Also ready to merge or need more stuff?

Constrat:
After EN gamedata drops we could release a beta. Let's hope CN gets their new gamedata otherwise a beta is unlikely.
cc @ABA2396 What do you think=

PR #12278, opened 2025-04-08
feat: 支持新 UI 主线导航

ABA2396:


ABA2396:
cc @Constrat Now the logic should be to avoid the original navigation, see if there is any effect

Constrat:
In 45 minutes, maintenance. Just in time

Constrat (approved):
Tested 13-11

ABA2396:
> In 45 minutes, maintenance. Just in time
> btw 
> {3548CB6C-9FC4-42E1-AC01-BA3ABDD70483}
> 

Is it too long? Then it should be good to add a max width

ABA2396:
image
I think this approach is a bit unconventional, but it actually works perfectly well.

PR #12348, opened 2025-04-13
Release v5.15.1

ABA2396:


Constrat (approved):
Resources updated.
Known bugs fixed. Nothing in the making for global (at least not known)

PR #12399, opened 2025-04-18
refactor: tasks拆分为文件夹

SherkeyXD:


src/MaaCore/Config/TaskData.cpp
```diff
@@ -218,6 +218,66 @@ bool asst::TaskData::lazy_parse(const json::value& json)
...
+                Log.error("TaskData::load", "Json open failed:", file);
+                continue;
+            }
+            json::value file_json = ret.value();
+            if (!file_json.is_object()) {
+                Log.error("TaskData::load", "Json content is not an object:", file);
+                continue;
+            }
+            for (auto& [key, value] : file_json.as_object()) {
+                merged.as_object()[key] = std::move(value);
```
SherkeyXD, line 255: 是否会影响加载速度？如果影响是否有 better pratice？

Constrat:
uuuuuuuuuuuuh how does this interface with all the global?

src/MaaCore/Config/TaskData.cpp
status102: 之前那个按模块拆分好像也是这条路子，应该只有一个任务重复的问题

SherkeyXD: 任务重复原本好像就有检查吧

status102: > 任务重复原本好像就有检查吧

之前走的应该是res-update-game.yml+TaskSorter.py，meojson好像是重复键不报错的

Constrat:
I'll look at the resource updater.

status102:
已知问题: res-update-game的original-json break了

Constrat:
This is actually worse than I imagined. Not only will the ResourceUpdater need heavy refactoring to the tasks to the correct files.json but also the workflow itself. Since it relies on single tasks.json with all keys in it to look through.

Constrat:
@SherkeyXD You did not move correctly the global stuff in roguelike: `Sami@Roguelike@FoldartalUseOcr` and `Sami@Roguelike@FoldartalGainOcr` were added by the updated ResourceUpdater

resource/tasks/Stages/SS.json
```diff
@@ -0,0 +1,189 @@
...
+        "action": "ClickSelf",
+        "postDelay": 2000,
+        "roi": [505, 57, 462, 628]
+    },
+    "SSSBattleOperName": {
+        "algorithm": "OcrDetect",
+        "text": [],
+        "fullMatch": true,
+        "roi": [0, 102, 190, 37]
+    },
```
Constrat, line 23: Is this supposed to be SSS.json or the base SS.json for the events?

SherkeyXD:
All the task breakdowns were done by my Python script, which doesn’t seem very smart lol.
I need to check these files again to see if there are any misclassifications.

resource/tasks/Stages/SS.json
Constrat: Resolved with [5982089](https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/12399/commits/59820893d3193b4f0c0ee752439283219a89afb2)

Constrat:
Ok i think the main.cpp should be ok for now. I will start working on the workflow itself.

Constrat:
{19EAFB81-6841-467A-B177-D352C73F8B00}
This is funny

SherkeyXD:
Are we ok now? Keeping this on will only cause more and more conflicts

Constrat:
Oh sorry. Forgot to undraft it. I thought you would since you were still working on refactoring no?

Constrat:
Btw, how would this interface with MaaRelease @SherkeyXD

Constrat:
Omg. Of course it's not working. Who called the Global folders `Stage` instead of `Stages`

Constrat:
Before merge. Does https://github.com/MaaAssistantArknights/MaaRelease need refactoring as well?
And going forward?

ABA2396:
照理来说先加载tasks再加载cache是没影响的，Release应该不用管

Constrat:
So for new events we still use the main tasks.json? And don't create singular XX.json with the data in it to delete later.

Constrat (approved):

Constrat:
Oh... Right... Hard to fix conflicts when we are moving the file away.

Constrat:
I'd suggest closing this one in favour of #12448

SherkeyXD:
This still causes conflicts —— as long as dev are making changes to tasks.json, we'll have to fix it from time to time. And it will also remove all commit history for tasks.json, which might be a problem in the future.
To merge as fast as possible might be the ultimate choice. Just let me solve the conflicts.

SherkeyXD:
You're right, squashing will solve the problem.
Let's move to #12448

PR #12498, opened 2025-05-01
fix: CheckLevelMax too fast without delay

BxFS:
For whatever reason current dev complains about ```Duplicate key in json file: MaaAssistantArknights\x64\Debug\resource\tasks\Stages\SS.json SS-Open``` on my environment, so I'm unable to test this myself. Shouldn't really be a problem as a one-liner tho.


ABA2396:
手动把 SS.json 删了

PR #12545, opened 2025-05-03
fix: 一键轮换模式自动补全必要设施参数

Lemon-miaow:
#12542


Constrat:
fix https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues/12542

PR #12556, opened 2025-05-04
fix: 基建选人前收起职业栏

ABA2396:

It's quite peculiar—recently, many users have reported that the dormitory pulls away operators who are working. It's unclear whether this was caused by a recent update or if it has always been an issue.

OnShift's minimization should be fine. Let's check if other clients can use this image. If they can, there's no need to reload the template.

For InfrastCloseQuickFormationExpandRole, we'll need to take a screenshot. But even if it's not collapsed, it shouldn't cause any issues—the minimized OnShift shouldn't fail to recognize it.
image
image
image


Constrat:
@ABA2396 I noticed that my 4 th dormitory (which MAA should skip since it has the skip option) loses the operators once every 2 / 3 days. My take on this is this stupid button here.

{FAA5FB34-A3DD-4D09-B809-4B97ACD6C2B1}
This is the worse addition ever added since it basically fucks up the rotations. I haven't really checked if MAA clicks it. But if I do click, it will remove all operators from dormitories that MAA should skip.

Constrat:
Regarding the PR issue tho, the chinese bar is much darker lol
image

ABA2396:
Previously, I had a headache over the transparency of the profession column in the copilot automatic formation, as it affected the recognition of profession icons. However, this works out fine—with a clear background distinction, simply collapsing this column resolves both issues seamlessly.

ABA2396:
> Oh yes definitely. I'll keep investingating the weird "remove all operators" from the last dormitory.
> 
> I swear this feature is so stupid 440170676-497d1c24-563d-4daf-8085-2b76e9ae5733

This feature currently requires enabling in the settings for the CN version. I recall there were restrictions implemented in the tasks—to prevent accidental clicks on this, there's also a waiting period for animations to finish after interacting with other elements.

Constrat:
Yeah I have a friend that plays on the CN and told me this will be disableable soon trademark

PR #12568, opened 2025-05-05
Release v5.16.2

ABA2396:


ABA2396:
「use medicine」does not trigger any adjustments. The dynamic modification only occurs when the medicine window is 「manual close」, because the system fundamentally only hooks the `CloseStonePage` event. The 「use medicine」button actually triggers `MedicineConfirm`.

ABA2396:
这下 Highlight 比正文多了

AnnAngela:
@Constrat i think we need to translate the flow, again

Constrat:
Make all the final changes. That way I'll just have to do it once XD

ABA2396:
> Make all the final changes. That way I'll just have to do it once XD

Ok. Let's do it according to the current version

ABA2396:
auto功能后续还要修改，文档暂时就不加了，看看之后要怎么改

PR #12573, opened 2025-05-05
rft: standardize Integrated Strategies numbers with prts.wiki

Constrat:
{C8ED6A40-98C8-4EB6-9878-11A82EA70777}
Keep consistency between official and codebase.
Majority of the users call IS4 sarkaz instead of sami so the codebase usually makes little sense.


Daydreamer114:
bro the official code is 4 for sarkaz, I've already told u when writing the seed code

Constrat:
🤔 uhm. The code as in the gamedata? I agree. But when colloquialy

PR #12602, opened 2025-05-08
feat: 添加 自定义 webhook 功能

KagurazakaIris:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/issues/9082
1. 添加了 自定义 webhook 功能。
2. 修复了部分UI文本中的错误。

目前经过在 Windows11 x86 上的简单测试，可使用 企业微信群机器人 的 webhook 发送消息。
image


Constrat:
@ABA2396 The `「」` at least in english are really not used that much, if ever actually. So the idea of removing all standard `""` for  `「」` is not that good. Especially since it adds weird spaces / padding whereas in english the space is much more important ` `

Constrat:
> In fact, `「」` is not commonly used in Chinese either, but comparatively, I'd prefer it not to be converted into `&quot;`.

Perfect. That gives me the reason why. That's all I needed. I know understand using standard quotes will definitely mess with the xaml.

ABA2396:
How about this? These are halfwidth corner brackets: `｢ (U+FF62)` `｣ (U+FF63)`.

ABA2396:
> Perfect. That gives me the reason why. That's all I needed. ~I know understand using standard quotes will definitely mess with the xaml.~ Actually will it tho? That xalm uses >< as separators no?

But the formatting doesn't think so
Think about the previous Weblate.

Constrat:
`｢｣`   `「」`. Oh yeah definitely. The first ones are much much better. No unnecessary spaces. Anyway as status102 said. We'll make these changes in a different PR definitely.

KagurazakaIris:
Thank you both for your valuable reviews. I appreciate your feedback and suggestions.

Regarding the formatting changes included in this PR, would it be preferable for me to remove them to maintain focus solely on the new webhook functionality?

I'm eager to see this feature integrated into the main branch and am open to making any necessary adjustments to facilitate a smoother merge process.

Constrat:
I haven't seen any formatting changes actually. pre-commit is not throwing any errors so this looks good. 

What we discussed about has already been fixed by status102 with a rebase.

ABA2396:
@KagurazakaIris 
image
你确定这个能用吗？转义需要处理吧

PR #12667, opened 2025-05-11
ci: don't check github link

wangl-cc:
不知道是 Github 的 Rate limit 更低了还是怎么回事。最近 check dead links 就没通过过，所以暂时不检查 github 的链接来避免这个问题。

### DIFF
--- a/.github/workflows/markdown-checker.yml
+++ b/.github/workflows/markdown-checker.yml
@@ -50,9 +50,6 @@ jobs:
             --verbose --no-progress
             --cache --max-cache-age 1d --cache-exclude-status '429, 500..599'
             --exclude 'https?://.*'
-            --include 'https?://github\.com/.*'
-            --include 'https?://raw\.githubusercontent\.com/.*'
-            --include 'https?://user-images\.githubusercontent\.com/.*'
             --exclude-path 'docs/zh-tw/manual/introduction/introduction_old.md'
             --exclude-path 'docs/ja-jp/manual/introduction/introduction_old.md'
             --exclude 'files/MAA_Runtime_Fix_Pwsh.ps1$'
--- a/.github/workflows/markdown-checker.yml
+++ b/.github/workflows/markdown-checker.yml
@@ -35,20 +35,12 @@ jobs:
         with:
           show-progress: false
 
-      - name: Cache Primes
-        id: cache-primes
-        uses: actions/cache@v4
-        with:
-          path: .lycheecache
-          key: lychee-cache
-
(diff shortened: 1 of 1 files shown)


Rbqwow:
好像...也行...?
这东西最开始好像就是为了检查内部超链接来着

以及让我想起来了原来我还有个 https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/docs/.vuepress/public/files/MAA_Runtime_Fix_Pwsh.ps1 的备份 好好好

Constrat:
@wangl-cc Me and @AnnAngela have been trying. Apparently... Either the token perms are not enough / it's not being used / or even with the token we are breaching over 1000 requests per hour.

To be honest, checking for github links doesn't really make sense in a way, since to make the links not work, the entire github backend needs to be down... Which is a much greater problem.

I agree in removing the github links from the checks.

Constrat:
I agree on the github side. I've tried using the GITHUB_TOKEN in every way possible. Modifying the permissions: to content true, using it in the args, passing it as env:, pasing it as token like the official lychee docs suggest, but it always hit the rate limiter :P

.github/workflows/markdown-checker.yml
```diff
@@ -35,24 +35,13 @@ jobs:
         with:
           show-progress: false
 
-      - name: Cache Primes
-        id: cache-primes
-        uses: actions/cache@v4
-        with:
-          path: .lycheecache
-          key: lychee-cache
-
```
AnnAngela, line 44: 缓存还是要保留的，能进一步减少请求数量

.github/workflows/markdown-checker.yml
```diff
@@ -35,24 +35,13 @@ jobs:
...
       - name: Check dead links
         uses: lycheeverse/lychee-action@v2
         with:
-          # 仅检查内部链接和 Github 相关链接
+          # 仅检查内部链接，排除所有外部链接
           args: >
             --verbose --no-progress
-            --cache --max-cache-age 1d --cache-exclude-status '429, 500..599'
```
AnnAngela, line 51: 同上

AnnAngela (changes requested):

.github/workflows/markdown-checker.yml
wangl-cc: 我直接删掉缓存其实是因为不检查 GitHub Link 之后就只检查内部链接了，不包含任何 http(s)，所以没有缓存的必要性。不过保留也行，如果之后要加上的话。

PR #12740, opened 2025-05-18
perf: 自动战斗编队优化干员识别, 修复roi未包含完整干员名

status102:
当前: 
D_T%GTU@ENGO0ENG)WYDT0M

修复后:
image


Constrat:
They started giving CN 7 characters? wow

status102:
> They started giving CN 7 characters? wow

already nearly 10 month ago

status102:
> This is a refactor. So will the image modification change? Because currently, at least for EN, everything works I have no idea how.

roi just extend to bottom and left. 
Main problem is wether cv related thing would broke overseas. works well on CN

Constrat:
So It's not just a refactor with no logic change. There were some changes

status102:
等待进一步分离

PR #12833, opened 2025-05-30
feat: RA store

Daydreamer114:


Constrat:
This could actually be run directly from the end of the normal RA farming taskchain? As when it ends automatically (because we got all the exp) we could go to the store.

Daydreamer114:
> This could actually be run directly from the end of the normal RA farming taskchain? As when it ends automatically (because we got all the exp) we could go to the store.

天才

PR #12895, opened 2025-06-05
chore: YostarJP Infrast edits

Manicsteiner:


Constrat:
@Manicsteiner I think we need also all new templates for all the themes.

(I didn't notice sami and the other one. We good with the other themes?)

Constrat:
Actually. All the mini facilities are broken... At least on EN. Everytime it exists and checks the notification it will die.

(Not just reception)

PR #12901, opened 2025-06-05
feat: EN base no beta

Constrat:


Constrat:
pre-commit.ci autofix

zayn7lie:
Base seems to be stuck on Notification
image

You can test it by leaving empty dorms with an op that morale is not full.

Constrat:
Yeah all templates need to be changed. Unfortunately yesterday yostar fucked up the login servers so I couldn't really test a lot. Basically try to have NO NOTIFICATIONS that way there's the bigger base preview. When it had the smaller rooms it will die since the base was increased in size

PR #12917, opened 2025-06-06
chore: feature-request增加陷阱选项 改进en-bug-report标题

Lemon-miaow:
~~我又来了~~


Daydreamer114:
Do not doubt or feel like lost
Everything's Alright with you!

PR #13045, opened 2025-06-20
fix: skip blank stages in KR replacement on ResoureUpdater

HX3N:


Constrat:
Did not approve. This is the YOLO achievement XD.

PR #13058, opened 2025-06-22
docs: 用户文档基建换班添加队列轮换说明

Lemon-miaow:


ABA2396:
i18n.jpg

Constrat:
One sided change. Please remember i18n. Otherwise the discrepancies will just keep increasing.
Even fully AI is good. Nowadays AI translation are almost too solid.

Lemon-miaow:
> One sided change. Please remember i18n. Otherwise the discrepancies will just keep increasing. Even fully AI is good. Nowadays AI translation are almost too solid.

That's really a good way to go about it. But since one problem is that the AI usually doesn't translate some of the game's proper nouns correctly, I'm a bit hesitant to use it.😰

Constrat:
You'd be surprised. If you tell the AI to think about Arknights terms. It's actually kind of accurate.

Lemon-miaow:
Got it. I'll give it a try

PR #13077, opened 2025-06-24
chore: IS4 invest with seed

Daydreamer114:
testing


Daydreamer114:
who removed the core logic

PR #13094, opened 2025-06-27
chore: 更新繁中服線索與部分截圖

momomochi987:
主要是更新線索的截圖，因為長不一樣了

其他幾個截圖只是順便 (x


resource/global/txwy/resource/template/InfrastSortByTrustButtonClickAgain.png
Constrat, line 1: Both templates seem lower resolution?

momomochi987: I have no idea why it seems like lower resolution...  😭😭😭
But it's working well on my computer 🤔

momomochi987:
@Constrat 

I think this pull request can be merged! 👀

If it causes any other issues, I’ll take a look later.

PR #13117, opened 2025-07-01
fix: 调整 YoStarEN 大荒城公招、抽卡的截图以修复识别失败的问题

gui-ying233:
小写改大写了…
（不让 Merge 于是重新开了个PR


Constrat:
No need for such specific templates. To keep templates standardized use the python script. Throw in the big 1080p image in the src and use the dst folder.

gui-ying233:
> To keep templates standardized use the python script. 

These two images were cropped by using the Python script and I don't think the ROIs need to be changed, am I missing something?

Constrat:
The templates are different size. The python script is the `GetImageFromROI`

Just to be clear. Merging like this will definitely work . It's just trying to keep consistency

Constrat (approved):
Lgtm. Thank you for the pr

gui-ying233:
> Lgtm. Thank you for the pr

Thanks for the help, seems the doc needs to be updated

Constrat:
The docs are fine for the most part. It's just for the main navigation we have a specific script

PR #13241, opened 2025-07-16
feat: 适配长期剿灭

dikxingmengya:
close: #4583 

## 外服适配（global adaptation）

因为新加入的功能，我们需要再次进行国际化适配。
Because we've added the new hit function, we need to conduct global adaptation again.

- [x] EN
- [x] KR
- [x] JP
- [x] txwy

请求对我提交的适配进行检查（EN,KR,JP）。
Request for checking  the adaptation I submitted (EN, KR, txwy).

- [ ] check

## 界面更改（MAA UI Change）

- [x] 已完成 / Finish

将三张长期剿灭和“当期剿灭”统一为“剿灭模式”，把选图功能放到“高级设置”中，以节省关卡选择中的列表长度。
Unify the three long-term Annihilation  and "Current Annihilation" into "Annihilation Mode", and put the selection function into "advanced settings" to save the list length.

### DIFF
--- a/src/MaaWpfGui/Res/Localizations/en-us.xaml
+++ b/src/MaaWpfGui/Res/Localizations/en-us.xaml
@@ -647,7 +647,7 @@ Do not adjust the proxy multiplier setting in the game</system:String>
     <system:String x:Key="PR-C-2">Pio/Sup Ⅱ</system:String>
     <system:String x:Key="PR-D-1">Grd/Spc Ⅰ</system:String>
     <system:String x:Key="PR-D-2">Grd/Spc Ⅱ</system:String>
-    <system:String x:Key="AnnihilationMode">AnnihilationMode</system:String>
+    <system:String x:Key="AnnihilationMode">Annihilation Mode</system:String>
     <system:String x:Key="Annihilation">Annihilation</system:String>
     <system:String x:Key="Chernobog">Chernobog</system:String>
     <system:String x:Key="LungmenOutskirts">Lungmen Outskirts</system:String>


resource/tasks/tasks.json
```diff
@@ -1150,7 +1150,7 @@
     "StartUp@ClickCorner": {
         "next": ["StartUp@ReturnButtons#next", "StartUp", "StartUp@CloseAnnos#next", "StartUp@OfflineConfirm"]
     },
-    
+
```
status102: 删删空行

src/MaaWpfGui/Res/Localizations/zh-cn.xaml
```diff
@@ -617,6 +617,9 @@ C:\\leidian\\LDPlayer9。\n
     <system:String x:Key="PR-D-1">近/特芯片</system:String>
     <system:String x:Key="PR-D-2">近/特芯片组</system:String>
     <system:String x:Key="Annihilation">当期剿灭</system:String>
+    <system:String x:Key="Chernobog">切尔诺伯格</system:String>
+    <system:String x:Key="LungmenOutskirts">龙门外环</system:String>
+    <system:String x:Key="LungmenDowntown">龙门市区</system:String>
```
status102, line 654: 给外服也加上

resource/tasks/tasks.json
dikxingmengya: 感谢

src/MaaWpfGui/Res/Localizations/zh-cn.xaml
dikxingmengya: 主要不知道外服的名字

Constrat: You've already added it for the key XD. That's actually the correct term, but don't worry. I will assemble the squad.
@MaaAssistantArknights/global-support-team 

Let's adapt before merging.

resource/tasks/tasks.json
```diff
@@ -1150,13 +1150,49 @@
...
+        "text": ["龙门市区"],
+        "action": "ClickSelf"
+    },
+    "AnnihilationSwitchButton": {
+        "Doc": "切换剿灭关卡（菜单按钮）",
+        "algorithm": "OcrDetect",
+        "text": ["切换"],
+        "action": "ClickSelf",
+        "next": ["#back"]
+    },
```
Constrat: Where is the `"roi":`? This is very heavy and will lead to wrong results without roi.

We surely don't need the full `0, 0, 1280, 720`

For the `Switch`, `1075, 661, 64, 36` is probably good enough:
{295CDA54-D741-4300-A6BF-DE5F5F1C7BC1}


For the annihilation stages we have two options. Either have a unified ROI like `892, 216, 262, 289`:
{1C82FB8E-A15D-4F66-A7E9-C505BA0B5748}

Or (I'm not sure) if they are ALWAYS in the same spot we can pinpoint the ROI much closer like this:
`887, 464, 156, 38`
{5FE58531-7649-402D-A5F4-963237ACD5CB}

dikxingmengya: I'm sorry but I don't know how to add the key "roi", I mean I don't know how to confirm the location.
And I had test it and it runs well, so I don't add it before.
I'll fix it soon.

dikxingmengya: I have added "roi" in CN

Constrat:
@HX3N It doesn't really matter but I'm curious. Why the custom ROI for KR? Is there a big difference?

Also @dikxingmengya this seems quite a big ROI... I think we can omit the annihilation "images" and perhaps the "arrows" with the "complete" indication
{7F16BE16-2C95-415B-A44E-5A5CDE3E2E07}

dikxingmengya:
> Also @dikxingmengya this seems quite a big ROI... I think we can omit the annihilation "images" and perhaps the "arrows" with the "complete" indication

@Constrat Because, these days the UI of Arknights have changed many times, so I think the big ROI could better adapt some small changes.
And the other reason is that I don't know the UI of the other server.
If there are some wrong with this in other server, you can change it.

HX3N:
> It doesn't really matter but I'm curious. Why the custom ROI for KR? Is there a big difference?

I just added it because it might be misrecognized, but after testing it doesn't seem to be needed

Constrat:
> the UI of Arknights have changed many times

I understand but trust me. If they change the UI again. We'll have to change roi anyway as it will be a big change xD

Anyway lgtm

dikxingmengya:
@Constrat We need to global adaptation again because of the file changes.Thanks.

dikxingmengya:
Request checking i18n before merging.

Constrat:
So I've added a space to keep consistency, but I don't think using this here is correct.
{2CF45A66-0517-407D-B025-A07545967ECF}

This is used in this instance which is correct as it allows users to choose their preferred annihilation mode in the more advanced settings.
{75851464-51B4-448B-962A-FDE70FCFBC66}


But in the main general settings I would leave the usual `Annihilation` anyway since we want to farm a stage... Not a mode.

DavidWang19 (approved):

dikxingmengya:
> So I've added a space to keep consistency, but I don't think using this here is correct. 
> This is used in this instance which is correct as it allows users to choose their preferred annihilation mode in the more advanced settings. 
> But in the main general settings I would leave the usual `Annihilation` anyway since we want to farm a stage... Not a mode.

I know, but actually this changes is requested by @ABA2396.

PR #13562, opened 2025-08-06
 fix: YostarKR update ocrReplace and ocr_config

HX3N:
<details><summary>log</summary>
스크린샷 2025-08-06 202024  
</details>

I thought it wasn’t being recognized because of a regular space, but it turned out to be an `&nbsp;` :\

### DIFF
--- a/resource/global/YoStarKR/resource/tasks/Roguelike/Mizuki.json
+++ b/resource/global/YoStarKR/resource/tasks/Roguelike/Mizuki.json
@@ -4,15 +4,7 @@
         "text": ["생태", "계보"]
     },
     "Mizuki@Roguelike@NextLevel": {
-        "ocrReplace": [
-            [".*하늘빛.?해안", "하늘빛해안"],
-            [".*연안.?외딴섬", "연안외딴섬"],
-            [".*맹렬한.?파도", "맹렬한파도"],
-            [".*은밀한.?온상", "은밀한온상"],
-            [".*신성한.?동굴", "신성한동굴"],
-            [".*심해의.?숲", "심해의숲"],
-            [".*짙푸른.?요람", "짙푸른요람"]
-        ],
+        "ocrReplace": [[" ", ""]],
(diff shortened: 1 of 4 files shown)


Constrat:
The OCR can detect &nbsp;... What the hell? Perhaps you can try the new IsAscii?
cc @ABA2396

HX3N:
Would it be better to just replace the nbsp?

---

> Perhaps you can try the new IsAscii?

Do you mean `"isAscii": true`? As far as I know, this doesn't work with Korean.
Has something changed, or are you referring to something else?

Constrat:
Not sure about other choices... But if you are sure I can change the resource updater to remove also that stuff.

HX3N:
>  But if you are sure I can change the resource updater to remove also that stuff.

Unfortunately, this seems to be an issue with the korean ocr model itself, so I don't think the resource updater can fix it.
For some reason, YostarKR's game(text) data contains a mix of Japanese, Chinese, and other languages, which causes the ocr model to recognize text in unpredictable ways. XP

ABA2396:
Here it's not a full-match. As long as it includes, it can be replaced. Therefore, the ".* "at the beginning and the end are unnecessary

ABA2396:
Can Korean actually recognize Spaces that are not ordinary Spaces? Should we exclude the dictionary

ABA2396:
You can add equivalent characters in `ocr_config.json`. Refer to the files in the jp folder for reference.

ABA2396:
image

Constrat:
Put all spaces as "identical" standard space ` `<-- This

HX3N:
> Can Korean actually recognize Spaces that are not ordinary Spaces? Should we exclude the dictionary
> You can add equivalent characters in `ocr_config.json`. Refer to the files in the jp folder for reference.

Yes, it can recognize non ordinary spaces as well.
The `ocr_config.json` makes those characters be recognized as the same character, right?

ABA2396:
Y

Constrat (approved):

PR #13623, opened 2025-08-10
fix: start plugin task not enough delay

Constrat:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/commit/4a68fac63975ec01f20b1292531f711365c4063e#r163749694

> This completely breaks the algorithm. 300 ms is too little to detect a crash. I remember during the "studies". 1500 is the minimum possible delay. Anything less and the game will not crash in time as it's still "booting up" or "running" right before the crash.

### DIFF
--- a/src/MaaCore/Task/Miscellaneous/StartGameTaskPlugin.cpp
+++ b/src/MaaCore/Task/Miscellaneous/StartGameTaskPlugin.cpp
@@ -18,8 +18,7 @@ bool StartGameTaskPlugin::start_game_with_retries(size_t pipe_data_size_limit, b
             }
         }
 
-        int delay = std::min(1500, 300 + i * 50);
-        sleep(delay);
+        sleep(1500);
     }
 
     return false;


ABA2396:
Is this part meant to avoid the game crashing right after launch?
I used to think it was simply checking whether the game had returned to running in the foreground.

ABA2396 (approved):

ABA2396:
But in this case, `newer_android` wouldn’t be checking for crashes, right?

Constrat:
So for whatever reason the yostaren client has the chance of returning to the main emulator main right after starting it. This happens both when the user manually clicks the application and when MAA uses the start intention to open the app. 

Before I added this small algorithm MAA would just open the game once and start all the tasks, but in reality there's a chance (big chance) for the game to go back to the main menu and MAA has no idea and would fail all the tasks.

Regarding the  newer version... I honestly forgot how it worked. Quite a bit of time has pssed...
Basically there's a better output for newer android that can tell us better what's happening so when the game correctly started / crashed. So we don't have guess and hope arknights started. We just check the output 


Perhaps if you are unsure we can just use this algorithm for yostaren only

PR #13624, opened 2025-08-11
fix: add delay before InfrastEnteredFlag

Constrat:
**UNTESTED** 
###### will check in the morning. 4AM currently.

I've noticed on my other machine that MAA keeps going back to the main interface because the score of the InfrastEnteredFlag is 0.82~0.86, when it's usually 0.96 (when correctly detected). This is likely because when going back from the trading post / power plant / dormitory there can be a bit of lag rerendering the base leading to a slight stutter, reducing the score. Just lowering could be a solution since before d65cde757482cd69e7090fccb5da4734ee1d73c7 there was the default 0.8 score. 

If for some reason the 0.86 score is still uninteractibile there shouldn't be big issues since we have a postDelay of 5000 anyway.

### DIFF
--- a/resource/tasks/tasks.json
+++ b/resource/tasks/tasks.json
@@ -2461,10 +2461,10 @@
         "next": [
             "Infrast",
             "InfrastNotification",
-            "InfrastPreEnteredFlag",
             "Infrast@ReturnButtons#next",
             "Infrast@CloseAnnos#next",
-            "Infrast@OfflineConfirm"
+            "Infrast@OfflineConfirm",
+            "InfrastPreEnteredFlag"
         ]
     },
     "Infrast@MainThemes": {
@@ -2491,7 +2491,7 @@
         "maskRange": [1, 255],
         "roi": [930, 540, 230, 180],
         "action": "ClickSelf",
-        "next": ["InfrastPreEnteredFlag", "Infrast", "Infrast@CloseAnnos#next", "Infrast@OfflineConfirm"],
+        "next": ["Infrast", "Infrast@CloseAnnos#next", "Infrast@OfflineConfirm", "InfrastPreEnteredFlag"],
         "postDelay": 5000,
         "crop_doc": {
             "roi": [935, 570, 85, 75],
@@ -2501,7 +2501,7 @@
     "InfrastPreEnteredFlag": {
         "algorithm": "JustReturn",
         "postDelay": 1000,
-        "next": ["InfrastEnteredFlag"]
+        "next": ["InfrastEnteredFlag", "Infrast#next"]
     },
     "InfrastEnteredFlag": {
         "template": "InfrastOverview.png",
--- a/resource/tasks/tasks.json
+++ b/resource/tasks/tasks.json
@@ -2461,7 +2461,7 @@
         "next": [
             "Infrast",
             "InfrastNotification",
-            "InfrastEnteredFlag",
+            "InfrastPreEnteredFlag",
             "Infrast@ReturnButtons#next",
             "Infrast@CloseAnnos#next",
             "Infrast@OfflineConfirm"
@@ -2491,12 +2491,17 @@
         "maskRange": [1, 255],
         "roi": [930, 540, 230, 180],
         "action": "ClickSelf",
-        "next": ["InfrastEnteredFlag", "Infrast", "Infrast@CloseAnnos#next", "Infrast@OfflineConfirm"],
+        "next": ["InfrastPreEnteredDelay", "Infrast", "Infrast@CloseAnnos#next", "Infrast@OfflineConfirm"],
         "postDelay": 5000,
         "crop_doc": {
             "roi": [935, 570, 85, 75],
             "mask": []
         }
+    },
+    "InfrastPreEnteredDelay": {
+        "algorithm": "JustReturn",
+        "postDelay": 1000,
+        "next": ["InfrastEnteredFlag"]
     },
     "InfrastEnteredFlag": {
         "template": "InfrastOverview.png",
--- a/resource/tasks/tasks.json
+++ b/resource/tasks/tasks.json
@@ -2491,14 +2491,14 @@
         "maskRange": [1, 255],
         "roi": [930, 540, 230, 180],
         "action": "ClickSelf",
-        "next": ["InfrastPreEnteredDelay", "Infrast", "Infrast@CloseAnnos#next", "Infrast@OfflineConfirm"],
+        "next": ["InfrastPreEnteredFlag", "Infrast", "Infrast@CloseAnnos#next", "Infrast@OfflineConfirm"],
         "postDelay": 5000,
         "crop_doc": {
             "roi": [935, 570, 85, 75],
             "mask": []
         }
     },
-    "InfrastPreEnteredDelay": {
+    "InfrastPreEnteredFlag": {
         "algorithm": "JustReturn",
         "postDelay": 1000,
         "next": ["InfrastEnteredFlag"]


ABA2396:
It’s not feasible to put `JustReturn` at the front of `next`, because this would cause the task to be entered directly without checking the subsequent `next` entries. `JustReturn` can only be placed at the end of `next`, and at most one is allowed.

Constrat:
Yeahi was very sleepy. I'll take a better look at this tiday

Constrat:
@ABA2396 So what's a good way to just add a delay? There's no good "algorithm" as a DoNothing sort, like the `Action`

I think a new type of delay is required in the task itself. A real `preDelay` as in the delay before recognition is done, since... Well, the `postDelay` of the previous task is really not scalable as it "delays" all the `next` tasks, instead of just the one you want.

ABA2396:
I think directly postdelay+1000 is the same. If you want to enter this task, all the others have to wait for it, or you can insert the #next of individual tasks in the next section of the task

Constrat:
As in?
```json
    "InfrastBegin": {
        "algorithm": "JustReturn",
		"postDelay": 1000,
        "next": [
            "Infrast",
            "InfrastNotification",
            "InfrastEnteredFlag",
            "Infrast@ReturnButtons#next",
            "Infrast@CloseAnnos#next",
            "Infrast@OfflineConfirm"
        ]
    },
```

I don't really get the point of using #next in `next`.

My objective was to pinpoint and add the 1000 delay JUST to `InfrastEnteredFlag`, instead of using postDelat for every other task such as Infrast and Notification.

I still think there's not having room in the delays. Pre and Post are very limiting, since you can't work on any pre algorithm delays per singular task. You need to have a pre task that will handle the delay, increasing the complexity and task number with no benefit.

Perhaps implementing a preRecognitionDelay?

ABA2396:
I don’t quite understand what you mean. **Predelay** is the time interval between detecting the element and executing the action, while **postdelay** is the time after finishing the action before detecting the next one. The detection of “next” is sequential — the previous one must be finished before the next can be detected. If you want to add a delay to the previous task, then all subsequent tasks will also have to wait for that time. In that case, what difference is there from simply adding it to the previous task’s postdelay?

ABA2396:
Or, as I said above, you can put the inner JustReturn at the end of InfrastBegin (Infeast should be changed in the same way — here I’m just using InfrastBegin as an example), because `"Infrast"`, `"InfrastNotification"`, `"Infrast@ReturnButtons#next"`, `"Infrast@CloseAnnos#next"`, and `"Infrast@OfflineConfirm"` all have matches, and there is no JustReturn, so it matches what I described above.

> It’s not feasible to put JustReturn at the front of next, because this would cause the task to be entered directly without checking the subsequent next entries. JustReturn can only be placed at the end of next, and at most one is allowed.

So you can place `InfrastPreEnteredFlag` after them, making next become:

`"Infrast","InfrastNotification","Infrast@ReturnButtons#next","Infrast@CloseAnnos#next","Infrast@OfflineConfirm","InfrastPreEnteredFlag"`.

Then, in `InfrastPreEnteredFlag`, add a postDelay, and change next to `"InfrastEnteredFlag", "Infrast#next"`. This way, if the first round matches something else, it won’t go into `InfrastPreEnteredFlag`. If it does go into `InfrastPreEnteredFlag`, it will wait one second before prioritizing the recognition of `InfrastEnteredFlag`. If that’s not found, it will then re-check the items in Infrast’s next.

ABA2396:
```json
"Infrast": {
    "next": ["Infrast", "Infrast@CloseAnnos#next", "Infrast@OfflineConfirm", "InfrastPreEnteredFlag"]
},

"InfrastPreEnteredFlag": {
    "algorithm": "JustReturn",
    "postDelay": 1000,
    "next": ["InfrastEnteredFlag", "Infrast#next"]
},
```

ABA2396:
Or, similar to the previous two tasks, use **#back**. In that case, you can use `Infrast@InfrastPreEnteredFlag`.

Constrat:
Oh... Reading this again made me understand... Yes you are right. postdelay in the "before next" is the same as a "prepredelay" in the next task if they are sequential. But in this specific case, since `"InfrastEnteredFlag"` is after both `"Infrast"` and `" InfrastNotification"` wouldn't that add the delay to those two as well?

```mermaid
graph LR
    A["InfrastBegin"] --> C["postDelay(1000)"]
    C --> D["Infrast"]
    D --> E["InfrastNotification"]
    E --> G["InfrastEnteredFlag"]
```
^ Everything is delayed by a 1000 with no micromanagement possible.

```mermaid
graph LR
    A["InfrastBegin"] --> C["Infrast"]
    C --> D["InfrastNotification"]
    D --> E["prePreDelay(1000)"]
    E --> G["InfrastEnteredFlag"]
```
^ Just the last task is delayed thanks to the prePreDelay. This is currently doable with what you suggest, but only if this is the last task.

Graphically this is what I meant. Which is basically what you suggested with the "normal logic".
I wanted to explain how "contorted" the logic needs to be to do a simple delay. Everything needs to be rearranged with a priority to push the delays down the chain, instead of allowing them in the tasks themselves increasing scalability.

Constrat:
> Or, similar to the previous two tasks, use **#back**. In that case, you can use `Infrast@InfrastPreEnteredFlag`.

Ah I see. Use Infrast as a baseTask and use #back to said baseTask. This is interesting!

Constrat:
Oh, same issue.
```py
[2025-08-11 14:47:33.942][ERR][Px1a84][Tx1a88] │││││ Infrast@ReturnConfirm->next has a not-final JustReturn task: InfrastPreEnteredFlag
```

ABA2396:
Remove `InfrastPreEnteredFlag` from InfrastBegin’s next. The next of `Infrast@MainThemes` contains `InfrastBegin#next`, and `Infrast` has `ReturnButton`, whose next includes `MainThemes#next`, so `InfrastPreEnteredFlag` from InfrastBegin’s JustReturn will be added. Maybe just adding a bit of postdelay is still the easiest solution :)

Constrat:
Yes but that adds the delay to every task at the start and I don't like that as it adds useless delay. Hence why the graphs explaining the prepredelay xD... If I wanted to implement this delay... Where should I look? Process Task?

ABA2396:
Makes sense — and with this, "Infrast@ReturnButtons#next" comes before InfrastPreEnteredFlag, so it just returns right away.

ABA2396:
Then it should be at the same level as `postdelay` and `predelay` maybe call it `preMatchDelay` XD

Constrat:
> Then it should be at the same level as `postdelay` and `predelay` maybe call it `preMatchDelay` XD

Yes basically. So you can precise delays in all the steps

ABA2396:
cc @status102 @Alan-Charred

Constrat:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/b26f6be3643ebd006c58f6e83d8db45f82063070/src/MaaCore/Task/ProcessTask.cpp#L114-L122

I tried working around here but wasn't able to get any delay. At least not in the logs

Constrat:
@ABA2396 I will push a "temp" fix with a postdelay at `InfrastBegin`.I will leave this open as a reminder XD

status102:
> As in?
> 
> ```json
>     "InfrastBegin": {
>         "algorithm": "JustReturn",
>         "postDelay": 1000,
>         "next": [
>             "Infrast",
>             "InfrastNotification",
>             "InfrastEnteredFlag",
>             "Infrast@ReturnButtons#next",
>             "Infrast@CloseAnnos#next",
>             "Infrast@OfflineConfirm"
>         ]
>     },
> ```
> 
> I don't really get the point of using #next in `next`.
> 
> My objective was to pinpoint and add the 1000 delay JUST to `InfrastEnteredFlag`, instead of using postDelat for every other task such as Infrast and Notification.
> 
> I still think there's not having room in th

If there is a `preMatchDelay` in `InfrastEnteredFlag`:
1. hitting `InfrastBegin`
2. postDelay 1000ms
3. Matching `Infrast` -> failed
4. Matching `InfrastNotification` -> failed
5. preMatchDelay `InfrastEnteredFlag` 1000 ms
6. Matching `InfrastNotification` -> failed
...

every retry would delay 1000ms here.

Constrat:
But the delay would be intrinsically in the EnteredFlag, so worst case scenario it has to retry. In the best case scenario there's delay just for EnteredFlag.

Infrast or infrast notification don't fail, there won't be any additional delay as the post delay is not necessary in infrast begin

Constrat:
@ABA2396 Apparently it's all a template issue... We were using the old one in the global. You switch the flag to only check for the `off` in the building mode.
Which can be used by all clients, at least for En it works fine with a 0.98 score
InfrastOverview

ABA2396:
该接口是半透明的。现在在最右边有一个额外的设施栏。当你向右拖动它时，它最终会落在中间，然后变得无法辨认

ABA2396:
This interface is semi-transparent. Now there is an additional column of facilities on the far right. When you drag it to the right, it will eventually land on the center and then become unrecognizable

ABA2396:
image
image

Constrat:
Yep, I get what you mean. The issue wasn't really moving too far. It was just a slightly 0.94 score, which is likely because the older EN template is too old and required "an update", but why update when we can just use the CN one :) #13801

PR #13707, opened 2025-08-15
Revert "ci: bump the github-actions group in /.github/workflows with 2 updates"

MistEO:
Reverts MaaAssistantArknights/MaaAssistantArknights#13648

fix https://github.com/MaaAssistantArknights/MaaAssistantArknights/actions/runs/16971563354/job/48141111569


Constrat:
Apparently there were breaking changes 🥶

PR #13812, opened 2025-08-19
ci: split codeql

Constrat:


Constrat:
{578ED1A9-8EE2-4392-8A05-F401428BB909}
This is also displayed in the previous runs without the changes. But the security tab shows something so I'm not really sure about that.

Constrat:
...I tried not to touch anything codeql related as I haven't dug deep in the documentation yet

SherkeyXD:
> {578ED1A9-8EE2-4392-8A05-F401428BB909} This is also displayed in the previous runs without the changes. But the security tab shows something so I'm not really sure about that.

This one doesn't matter as we don't need to run it everytime someone pushes, the pr trigger is enough.

Constrat:
> This one doesn't matter as we don't need to run it everytime someone pushes, the pr trigger is enough.

Ah ok, when you have time let me now about the other changes

SherkeyXD:
Actually we are planning a refactor to the website and docs —— the website will be moved back to [maa-website](https://github.com/MaaAssistantArknights/maa-website) and the docs will be deployed to `docs.maa.plus`.
In this situation, we don't need `codeql-js` anymore, since the doc site is static and won't cause secutity problems, and for website we can do it at maa-website.

Constrat:
Ah sure. Welp I just have to remove the the first job. That's all and leave the GitHub actions

Constrat:
When the refactors of the website / docs comes, I'll make the necessary changes. For now this also bring cache fixes that speeds this up

PR #14022, opened 2025-09-07
feat: core 崩溃后下次启动时 ui 输出提示

ABA2396:

[crash.log](https://github.com/user-attachments/files/22193905/crash.log)
Set this as readonly and place it in the root directory of maa


ABA2396:
image

PR #14038, opened 2025-09-08
chore: prts -> zoot

Constrat:
Website was changed correctly, but MAA still refs to prts instead of zoot. It will work nonetheless as the routing is doing a great job, but it's still inconsistent.

Opinions?


Rbqwow:
well, it's just an easter egg due to the game story i think
and someday we'll change it back
cc @MistEO

MistEO:
Yes, Zoot is just a temporary 302 redirect.

PR #14064, opened 2025-09-10
chore: 繁中服「大荒」介面主題

momomochi987:


Constrat:
Interesting cmake error.
Why in the hell is the workflow running from `ci/smoke-test` and not from `dev`

PR #14074, opened 2025-09-11
chore: tweak workflow + remove .sln

Constrat:


Constrat:
Fuck, we need the .sln still. And here I thought we could just use the lists

PR #14076, opened 2025-09-11
feat: convert ResourceUpdater to ubuntu-latest

Constrat:
Finally, after two years. We have almost all workflows running in Ubuntu!
Speed!


.github/workflows/res-update-game.yml
```diff
@@ -140,55 +140,73 @@ jobs:
...
-          cmake --build build --target ResourceUpdater --config Release --parallel $env:NUMBER_OF_PROCESSORS
+          cmake --build build --config Release --parallel $(nproc)
 
-      - name: Install Resource Updater
-        if: steps.cache-resupd.outputs.cache-hit != 'true'
+      - name: Install
+        if: steps.resupd-cache.outputs.cache-hit != 'true'
         run: |
-          cmake --install build --component ResourceUpdater --config Release --prefix tools/ResourceUpdater
+          cmake --install build --config Release --prefix tools/ResourceUpdater
```
Constrat, line 197: @neko-para I had to remove both `--target ResourceUpdater` and `--component ResourceUpdater`.

In the install it was pretty simple why. It required the header files AsstType etc, but the build. It said it neede the main MaaCore library which by default is not made when building ResourceUpdater? Am I missing something?

neko-para: Not sure what you means... If doesn't specified target, then both MaaCore and ResourceUpdater shall be built. Is there any error occured?

Constrat: Sorry, I've selected the wrong lines.

Constrat, line 197: @neko-para Here. (I selected the wrong lines before)
I wanted to ask. Is there really no alternative? I can't remember if Windows was doing the same with the .sln
But I do remember that I was just caching the .exe and the opencv.dll, but not the DL / MaaCore ones. So perhaps ... There's a way?

Don't get me wrong. It doesn't really matter since the cache will do 99% of the runs.

Changes are usually required for new integrated strategies on global, or weird changes on yostar where operators don't appear on the recruitment. It's just curiosity basically.

neko-para: So, what indeed do you want? I just not get it. What's your expectation, and what's the situation now?

Constrat: Sorry. We were using these before for Windows. And it just 
`cmake -B build -DCMAKE_BUILD_TYPE=Release -DMAADEPS_TRIPLET='maa-x64-windows' -DBUILD_RES_UPDATER=ON`
`cmake --build build --target ResourceUpdater --config Release --parallel $env:NUMBER_OF_PROCESSORS`
`cmake --install build --component ResourceUpdater --config Release --prefix tools/ResourceUpdater`

{B2D0ACCC-73D4-49AE-83E0-DC99BBE88FC2}
{A16A0471-F6F3-4C25-8AC4-C1F125423F75}

_After that, we only need the `opencv`_

TL;DR is there a way to get the same results in Linux?

neko-para: Just want to build and install ResUpdater and opencv (and libcxx on linux for sure)? If so, just add the missing `--target ResourceUpdater` and `--component ResourceUpdater` shall work 🤔

neko-para: 61265768e5cc7cf037fccba9a2bb9a34481e12c9

I've add this commit to provide a standalone component for libcxx.

```shell
$ cmake -B build -DCMAKE_BUILD_TYPE=Release -DMAADEPS_TRIPLET='maa-x64-linux' -DBUILD_RES_UPDATER=ON \
-DCMAKE_TOOLCHAIN_FILE=cmake/linux/maa-x64-linux-toolchain.cmake
$ cmake --build build --config Release --target ResourceUpdater
$ cmake --install build --prefix tools/ResourceUpdater --component ResourceUpdater
$ cmake --install build --prefix tools/ResourceUpdater --component libcxx
$ ls tools/ResourceUpdater
clone_data_repo.ps1  clone_data_repo.sh  libc++abi.so.1  libc++.so.1  libfastdeploy_ppocr.so
libonnxruntime.so.1  libopencv_world4.so.411  libunwind.so.1  main.cpp  ResourceUpdater  version.ps1  version.sh
```

Constrat: > Just want to build and install ResUpdater and opencv (and libcxx on linux for sure)? If so, just add the missing `--target ResourceUpdater` and `--component ResourceUpdater` shall work 🤔

Unfortunately not really. I had to remove them in this PR because it wasn't working otherwise on the Ubuntu runner. Sorry for the delay, I was asleep

neko-para: But that shell output is on an ubuntu... So give it a try?

Constrat: Sorry. I meant before the libcxx. I was just answering to the first message.


Aka the changes made in this PR only. I'll make sure to have a look at the new suggested changes ASAP. Thank you

EDIT:

{D8B44DEC-96EA-402E-9D2E-2BDE0894F7AF}

Here, this is by only using the --target and --component without the new changes, just to give you an idea. 

I will now try with the new changes you suggested

EDIT2: 
Works perfectly!
{980779F1-D3B5-4A78-B6C6-364012B33D5B}

neko-para:
image

image

Newly cross compiled llvm, 276M after extraction.

Constrat:
Welp. The less the better! As we can more easily cache, but most importantly, more easily extract it I guess? Less files = less extraction time. Now we just need to " install it" in the right places?

neko-para:
Yes, I'm making an release for it. This is intended to be downloaded by `download-linux-toolchain.py`.

Constrat:
Does it go in MaaDeps? Because if it doesn't then it becomes a bit complicated to manage the steps.

neko-para:
MaaDeps contains the same script...

Actually, the reason why I add this script to Maa repo is that the toolchain files and downloading scripts are not distributed. MaaFramework is using submodule thus it can directly use them, but Maa isn't that case.

I'll add the toolchain file and downloading scripts to the distributed files of MaaDeps, as decided before.

Constrat:
Sorry. I wasn't clear. I meant where would llvm be placed / "installed" to. Will it be automatic in the python script? I ask this for caching purposes, because currently we cache the entire `MaaDeps` folder after running the maadeps and the toolchain scripts.

neko-para:
Yes, it'll be put under MaaDeps/x-tools/llvm, so that the distributed toolchain file can properly find it.

Constrat:
Perfect! That way we have a single "cache"

neko-para:
https://github.com/MaaAssistantArknights/MaaDeps/pull/34

MistEO:
Why don't we use macos-latest? This runner uses the M1 chip and will run faster.

MistEO:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/actions/runs/17702947996

For example, in this pipeline, Ubuntu build takes 6 minutes, while macOS (arm64) only takes 2 minutes. If we use macos for smoke testing, the speed will be significantly improved

Constrat:
Oh interesting. 
Well I guess sky's the limit @MistEO.
I actually never thought about MacOS for compatibility reasons but you're right... MaaCore wise there's no difference between Ubuntu and the M1 runners.

I'll make changes asap, or, you're free to do them yourself.

PR #14081, opened 2025-09-12
fix: 移除 debug_demo (原Sample) 的对应 resource 额外查找逻辑

status102:


Constrat:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/actions/runs/17671352535/job/50223655620

✅ Good run.
But what happens for cached binaries? Why do we need links when we use directly the main `resource` folder anyway?

status102:
> Why do we need links when we use directly the main resource folder anyway?

Because there would be a resource dir in the final running envirnment of users, it seems that is a little ugly to hard code a additional resource dir path for Debug env. So I add links in ci

Constrat:
You think? I mean it's a debug version so anything goes no?

status102:
> You think? I mean it's a debug version so anything goes no?

So it's just a PR :(

Constrat:
Sorry? I don't quite understand.

status102:
> Sorry? I don't quite understand.

- #13908 remove the `ResSync` (cpoy `resource` dir from project to output) then MistEO add a fallback path `../../../resource` for `Wpf` and `Sample`.

- #14077, #14072 Then uye and I add mklink to `Wpf.csproj` instead of `../../../resource`

this PR is to do same thing for `Sample`

CMakeLists.txt
```diff
@@ -124,6 +154,7 @@ if (BUILD_SMOKE_TEST)
     target_link_libraries(smoke_test MaaCore)
 
     add_dependencies(smoke_test MaaCore)
+
```
status102:
```suggestion
```

.github/workflows/smoke-testing.yml
```diff
@@ -112,6 +112,11 @@ jobs:
         run: |
           mkdir -p install
           cmake --install build --prefix install --config Debug
+          # 创建资源目录链接
+          echo "Creating resource directory link..."
+          rm -rf install/resource
+          ln -s "$(pwd)/resource" install/resource
```
MistEO: 已经有了为啥要删掉换link

status102: > 已经有了为啥要删掉换link

已经有了么，不过rm确实多余

MistEO: -DINSTALL_RESOUCE 之后再 cmake install 就会有

Constrat:
~~To be completely fair, I still don't understand why we have to install resources (copying them over to another location) when by default it will check the main location anyway without the need of duplication overhead, especially in a debug environment. This could also lead to local environment testing where you have to make changes twice both in the /resource and in the /install/resource to make sure the changes you make are real.~~

~~Having a centralised resource location seems best. It's a debug environment anyway.
Plus what happens when the smoke tester is cached? Will it check the /resource? Or will it try to check /install/resource which doesn't exist as the cmake --install doesn't run (and shouldn't run if you ask me, as it's just overhead, again)~~

~~It feels like we are slowing down something on purpose, for basically no reason / functionality at all.~~

~~Or did I completely miss the point of what this does~~

~~EDIT:~~
~~https://github.com/MaaAssistantArknights/MaaAssistantArknights/actions/runs/17696636415/job/50296950675~~

~~Exactly as I suspected. Now it will fail has it has no link to the main resources we clone. By link I don't mean the `ln` but the logical code to check /resource generically.~~

~~Ironically, this is the opposite of a `fix`.~~

I guess I can just move where we install the debug. That way we access the /resource anyway?

EDIT2: Fixed this by not using the install folder, but putting everything in ./

status102:
> I guess I can just move where we install the debug. That way we access the /resource anyway?
> 
> EDIT2: Fixed this by not using the install folder, but putting everything in ./

Not a problem. Install in ./ or add mklink after cache download.

PR #14096, opened 2025-09-13
ci: 为smoke-testing使用cache时增加resource链接

status102:


Constrat:
I still don't understand why. What's the difference? This just adds overhead for the linking step which is completely useless.
Is it a more similar testing environment for users? Using /install/resource instead of /resource? To check if the -DINSTALL_RESOURCES works correctly?

status102:
I just don't want to pile all the files together. And the overhead does not seem to be a big deal for actions in my opinion. and sry for bothering you.

status102:
> To check if the -DINSTALL_RESOURCES works correctly?

Btw, we needn't DINSTALL_RESOURCES now

Constrat:
No bother at all. I just wanted to make sure we needed it that's all.

We are team for a reason. If you bring me a good enough reason to make changes, I'll gladly accept them don't get me wrong. But this seems a bit "useless" especially `ci` wise?

Regarding the piling thing... It's an automated runner? It doesn't really care.

Yes you are right. We don't need the -DINSTALL_RESOURCES anymore as we use /resource directly. I'll remove it. Thank you.

status102:
Using of resource:
1. Wpf & Core: /build/bin/Debug/resource, mklink by msproj post build event
2. Debug_Demo with VS(sln, ASST_DEBUG):
	- If build Wpf there would be a link by msproj 
	- Run Demo independently, use ci to mklink
	So we needn't ../../../resource here
3. Debug_Demo for smoke-testing:
    - Choose 1: install in root dir
    - Choose 2: mklink to /install
    We have these way to use /resource path

But, Choose 1 mixes output files with source files. So I prefer latter

Constrat:
I mean I get what you mean, but personally a workflow needs to be efficient? The runner doesn't care about where files are. If you're fine with the way the main.cpp is now then there's nothing else to do with the workflow. As you can see I have to cache them by files and by folder anyway because before we were linking the resource with lead /install/resource and obviously we don't want to cache the entire /resource

status102:
I'm not sure if it would cache entire /resource. seems cache specific files, so it may not cache entire resource I guess?

Constrat:
I didn't explain myself well. Putting the built files in a single folder changes basically nothing as I still have to define the files singlehandedly because of the resources if we place them in "custom" non root folder

MistEO:
I discussed with @status102  that installing in the root directory is unacceptable in my opinion, as it will cause unimaginable troubles later. We need a clean and complete test directory to run tests, not mixed with a bunch of unrelated files

PR #14104, opened 2025-09-14
ci: try macos for smoke_test

MistEO:


Constrat:
~~Identical times to ubuntu-latest, perhaps even slower on the build times...~~
~~Now, what would be the difference between the building of ci.yml mac and this one~~

It's stuck at 98%
Plus it's not using parallelism

{27607994-5F92-4982-A71A-7640010F410D}

Not sure if macos has the flag CMAKE_BUILD_PARALLEL_LEVEL set by default, but the `nproc` equivalent is `sysctl -n hw.logicalcpu`

Constrat:
Damn, it doesn't like the shell script seriously?
{E74F9731-3455-49EE-BD6C-F4FD1894EC4A}

But build times are promising yes!
Let's hope the brew install doesn't add that much overhead.
I'll switch to a zsh script in case since they are basically identical... zsh is easier to work with actually.

Ok the bash install is slow af.

(The resource install is slow as well, plus it won't work when caching since it will install only on build)

MistEO:
update bash tasks about 20s, it seems that switching to zsh is a better choice, but the script already works, so why does it fail?

Constrat:
Feel like bash differences... Even though it makes no sense. `bash` is `bash`

Btw M1 is stupidly fast... Time to switch everything to M1 runners! 🤣 

Fuck perms.

neko-para:
-A is a bash extension 😆

Constrat:
typeset v declare.

Constrat:
Oh no. It broke the smoke testing now. 😨

MistEO:
my fault, I force pushed it

Constrat:
Building now 👍🏻

Constrat:
Do we fix the cached version missing the resources here as well?

Since when cache is hit it will skip all builds / install, meaning no /install/resource.
Solutions?
- Manual link like status102 suggested? 
- Reimplement the hardcoded /resource in ASST_DEBUG?

Fuck. I turned off wsl today. I'll have to restart it.
{904C47C5-519A-4A0E-B43F-3CC8A8039BA9}


I'll try with my tablet. Give me a sec.

MistEO:
When the cache is hit, we create a resource link, just like status102 does, how about that?

Constrat:
Sure. You said using /install is inevitable (and we don't want to add the ASST_DEBUG logic to use /resource as a fallback) so I don't really have other solutions other than making the links

Constrat:
Ok. The script should be fixed. For testing here. I'll enable caching temporarily in branches as well.

Constrat:
> libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument

Well this is not good.

Constrat:
Now it passed? What the hell.

@MistEO We can merge I guess. I'll keep monitoring to see if this comes up again.

Constrat:
libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
AAAAAAAA
It's random or something XD.

Let me simplify the zsh script.
I'm not sure it's the script tho.

MistEO:
look good to me, after you are ok, you can merge 🚀

Constrat:
There's that libc++abi error that comes out sometimes and I don't understand why.

MistEO:
I think this is a mistake in smoke_test itself, there seems to be a lock conflict somewhere

Constrat:
I never noticed that in windows / ubuntu. Perhaps a mac weird? I guess that could be annoying.

PR #14106, opened 2025-09-14
ci: refactor res_updater cmake

MistEO:


Constrat:
We'll also need a zsh script, like before xD

Constrat:
Do you think it's also faster for the actions/checkout steps?

Constrat:
Uhm I thought I pushed the chmod +x... Guess I didn't

Constrat:
Oh :O
Missing update gitignore.

Constrat (approved):

Constrat (approved):
Green workflow. script works well.I just hope the oxipng works well on macos

The M1 runner is stupidly fast.

Give me a sec. I'll test the 3 jobs before to see if there's an increase in extraction with the M1 processor.

MistEO:
> Do you think it's also faster for the actions/checkout steps?

I guess there won't be much difference. There are fewer macos machines, and sometimes they have to queue up. We only need to modify the time-consuming jobs.

Constrat:
Yep, you are absolutely right. It was queuing them up. Basically killing the parallel jobs.

PR #14113, opened 2025-09-15
ci: use download-artifact v4

MistEO:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/actions/runs/17725188293

boom again :(

### DIFF
--- a/.github/workflows/release-ota.yml
+++ b/.github/workflows/release-ota.yml
@@ -134,6 +134,15 @@ jobs:
           prerelease: ${{ fromJSON(needs.create-tag.outputs.prerelease) }}
           overwrite: true
 
+      - name: Create issue if failed
+        if: failure()
+        uses: actions-cool/issues-helper@v3
+        with:
+          actions: "create-issue"
+          title: "Failed make release OTA for Windows"
+          body: |
+            ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+
   make-ota-mac:
     needs: create-tag
     runs-on: macos-14
@@ -177,6 +186,15 @@ jobs:
           prerelease: ${{ fromJSON(needs.create-tag.outputs.prerelease) }}
           overwrite: true
 
+      - name: Create issue if failed
+        if: failure()
+        uses: actions-cool/issues-helper@v3
+        with:
+          actions: "create-issue"
+          title: "Failed make release OTA for macos"
+          body: |
+            ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+            
   release:
     needs:
       - make-ota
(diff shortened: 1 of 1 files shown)


Constrat:
🤣

Constrat:
I'll try to fix this to make use of artifact v5. Testing the ota workflow is a pain as I'm always scared of pushing a tag or a release even if I'm not in `dev` or `master`.

We will have to move to v5 sooner or later

Constrat:
I can see the files is correctly uploaded (of course it's using upload v4)

I really don't understand the issue. The file size is identical, and it's correctly downloaded in the `make-ota` step

Isn't this because it's missing the name? and probably the path?

 Starting download of artifact to: /home/runner/work/MaaAssistantArknights/MaaAssistantArknights

It doesn't tell the `ls` fuck

Constrat:
> V4: Previously, single artifact downloads behaved differently depending on how you specified the artifact:

		By name: name: my-artifact → extracted to path/ (direct)
		By ID: artifact-ids: 12345 → extracted to path/my-artifact/ (nested)
> V5: Now both methods are consistent:

		By name: name: my-artifact → extracted to path/ (unchanged)
		By ID: artifact-ids: 12345 → extracted to path/ (updated - now direct)




Since didn't put a name I'm guessing it defaulted to id meaning it made a nested folder with the name of the artifact itself, meaning we go in MaaReleaseConfig, when the file is probably one directory above?

Constrat:
Before with v4:
Starting download of artifact to: /home/runner/work/MaaAssistantArknights/MaaAssistantArknights/MaaReleaseConfig

Now with v5:
Starting download of artifact to: /home/runner/work/MaaAssistantArknights/MaaAssistantArknights

Yes confirmed.

Constrat:
What would you like to do @MistEO ?

Just modify this line:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/cfd463a7fd6b29f1414daaed556f97afc4c2fb9d/.github/workflows/release-ota.yml#L121-L124

to:

```yaml
      - name: Build OTA
        run: |
          cd build-ota
          $GITHUB_WORKSPACE/MaaAssistantArknights/tools/OTAPacker/build.sh 'MaaAssistantArknights/MaaAssistantArknights' ../config ${{ matrix.target }} 'MaaAssistantArknights/MaaRelease'

```
<hr>

Or this:
https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/cfd463a7fd6b29f1414daaed556f97afc4c2fb9d/.github/workflows/release-ota.yml#L96-L97
To this:
```yaml
      - name: Download release config
        uses: actions/download-artifact@v5
        with:
          name: MaaReleaseConfig
          path: ./MaaReleaseConfig
```

PR #14378, opened 2025-10-12
ci: Add repo owner judgment to some steps of `website-workflow.yml`

weinibuliu:


Constrat:
~~I understand the intent but I've checked your fork and your website-workflow (the one that gave you an error is pretty old) as we've removed the `actions/configure-pages@v5` a while back.~~

Nevermind. Apparently I'm blind. Perhaps instead of skipping the entire workflow. We can just skip the annoying parts.

weinibuliu:
Sorry, I checked again the [dev branch file](https://github.com/weinibuliu/MaaAssistantArknights/blob/dev/.github/workflows/website-workflow.yml), and found the `actions/configure-pages@v5` on https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/dev/.github/workflows/website-workflow.yml#L37-L38 . Besides, I checked again my branch, and trusted it is already newest. Could you give me more information?

I think checking the repository owner would help prevent the workflow from running on forks where it shouldn’t. This is why I open this PR.

Constrat:
Absolutely. I've edited my comment (I was clearly wrong [and blind])

Constrat:
https://github.com/Constrat/MaaAssistantArknights/actions/runs/18443352853/job/52546687469

This is my suggestion. It will still check for pnpm errors etc. But it will skip the github pages section. This should work in all forks (where actions are enabled obviously)

.github/workflows/website-workflow.yml
```diff
@@ -26,6 +26,7 @@ concurrency:
 
 jobs:
   build:
+    if: github.repository_owner == 'MaaAssistantArknights'
```
Constrat:
```suggestion
```

weinibuliu:
Never mind.

I think your fork is good. Could we merge it into dev branch?

weinibuliu:
@Constrat Well, it was merged. But I think [this step](https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/14378/files#diff-6d0d215ffc8ce2b20acb2aad9319b46ec4cd6c6094e7af061e58fe612732d4cdR67-R73) should be skipped on forks too.

Could you push a simple commit to fix this?

Constrat:
Uhm, to reach that step one would have to work on the `master` branch which we are extremely against or would have to toggle the Deploy to Prod, which is also something weird a user could do, but I can see the "Let's do that just in case"

Constrat:
Done with [8ec0974](https://github.com/MaaAssistantArknights/MaaAssistantArknights/commit/8ec0974d1b8339019cd202484ade6c51981f753d)

PR #14708, opened 2025-11-11
perf: 更新 243 极限效率一天三换排班表（20251111 修订）

Zsm1n:
## Summary by Sourcery

Enhancements:
- Revise the JSON configuration for the 243 layout to match the 20251111 update


Constrat:
Yeah the 243 3 rotations is ancient since it still uses the full Sui combo.
But from a MAA point of view it's kind of useless having a 3 rotation, since as you said it requires a manual rotation for Fiammetta and the dormitories.

Our main infrastructure contributor `bodayw` stopped so if you are interested since you've already updated the 243 3 rotations.

The 243 4 rotations is "already prepared". Me and bodayw optimized it so that MAA can directly choose the operators instead of having the auto fill dormitories (which could lead to issues in the long run).

If you don't mind having a look at that one as well since you've already modified this one.

<hr>

I'll be merging this one for now

Zsm1n:
> Yeah the 243 3 rotations is ancient since it still uses the full Sui combo. But from a MAA point of view it's kind of useless having a 3 rotation, since as you said it requires a manual rotation for Fiammetta and the dormitories.
> 
> Our main infrastructure contributor `bodayw` stopped so if you are interested since you've already updated the 243 3 rotations.
> 
> The 243 4 rotations is "already prepared". Me and bodayw optimized it so that MAA can directly choose the operators instead of having the auto fill dormitories (which could lead to issues in the long run).
> 
> If you don't mind having a look at that one as well since you've already modified this one.
> 
> I'll be merging this one for now

I understand that fixing the dormitory operators will reduce potential issues, but users who need operators for processing, Training Room assistance that requires mood recovery, or dormitory trust farming will all be unable to use this.

Zsm1n:
As you said , it requires a manual rotation for Fiammetta and the dormitories.So I also made a plan that doesn't need Fiammetta, but instead, the efficiency will be lower

PR #14785, opened 2025-11-17
ci: Simplify resource update workflow

AnnAngela:
Removed unnecessary delay and re-download steps for resources.

## Summary by Sourcery

Simplify the resource update CI workflow by eliminating explicit sleep and re-download steps and using a centralized wait-for-jobs action to synchronize artifact downloads

CI:
- Replace manual delay and retry logic for resource artifact downloads with a single wait-for-jobs step
- Remove commented out job dependencies and unnecessary re-download steps for txwy, Overseas, and Official artifacts

### DIFF
--- a/.github/workflows/res-update-game.yml
+++ b/.github/workflows/res-update-game.yml
@@ -210,21 +210,18 @@ jobs:
 
       - name: Download txwy
         uses: actions/download-artifact@v6
-        continue-on-error: true
         with:
           name: txwy
           path: ./tools/ResourceUpdater/Overseas/tw/gamedata/excel
 
       - name: Download Overseas
         uses: actions/download-artifact@v6
-        continue-on-error: true
         with:
           name: overseas
           path: ./tools/ResourceUpdater/Overseas
 
       - name: Download Official
         uses: actions/download-artifact@v6
-        continue-on-error: true
         with:
           name: official
           path: ./tools/ResourceUpdater/Official


.github/workflows/res-update-game.yml
```diff
@@ -201,61 +198,29 @@ jobs:
...
-      - name: Download txwy
-        id: download-txwy
-        uses: actions/download-artifact@v6
-        continue-on-error: true
+      - name: Wait for resource cloning
+        uses: yogeshlonkar/wait-for-jobs@v0
         with:
-          name: txwy
-          path: ./tools/ResourceUpdater/Overseas/tw/gamedata/excel
+          ignore-skipped: 'false'
```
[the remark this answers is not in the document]
AnnAngela: 你再想想.jpg

Constrat:
Oh wow. This is old as well... I remember I was looking for such an action and I thought I found it but it was just an enhanced "delay"

EDIT: just checked the readme. It appears VERY promising!

.github/workflows/res-update-game.yml
```diff
@@ -201,61 +198,29 @@ jobs:
...
+          jobs: |
+            Download Official Resources
+            Download Overseas Resources
+            Download Taiwan Resources
```
Constrat, line 208:
```suggestion
            Download txwy
            Download Overseas
            Download Official
```

These names do not coincide with the `name:` of the download steps no?


From the README:

>     # Comma or newline separated list of names of job dependencies for this step, it must be `name:` property of job if set
>     # required: true
>     jobs: 'job1,job2,Job 3'


I'm not sure why the developer uses this format for the jobs (it seems to show they can be generic? So case insensitive and spaces can be omitted?

Constrat (changes requested):

.github/workflows/res-update-game.yml
AnnAngela: I think these should match the `name:` of the jobs

AnnAngela: <https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/adb43c9736a7c73e99a12671975b0f8fb06263c4/.github/workflows/res-update-game.yml#L15>

<https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/adb43c9736a7c73e99a12671975b0f8fb06263c4/.github/workflows/res-update-game.yml#L47>

<https://github.com/MaaAssistantArknights/MaaAssistantArknights/blob/adb43c9736a7c73e99a12671975b0f8fb06263c4/.github/workflows/res-update-game.yml#L90>

AnnAngela: > From the README:
> 
> > ```
> > # Comma or newline separated list of names of job dependencies for this step, it must be `name:` property of job if set
> > # required: true
> > jobs: 'job1,job2,Job 3'
> > ```
> 
> I'm not sure why the developer uses this format for the jobs (it seems to show they can be generic? So case insensitive and spaces can be omitted?

I read the source code, so this action use `name` to match the jobs (`job.name === name`):

<https://github.com/yogeshlonkar/wait-for-jobs/blob/708b4037b51e860d3d4dbb1f0ab098e9a9dbcd80/src/WaitForJobs.ts#L86-L94>

> ```typescript
>     private toCheck = (name: string, jobs: Job[]): Dependency => {
>         const js = jobs.filter(
>             job =>
>                 (this.asSuffix && job.name.endsWith(name)) ||
>                 (this.asPrefix && job.name.startsWith(name)) ||
>                 job.name === name
>         );
>         return new Dependency(name, js, this.ignoreSkipped);
>     };
> ```

AnnAngela: And since they use `===`, then the `name` must be exactly the same.

Constrat: I was reading the wrong lines of code. My failt

PR #14820, opened 2025-11-21
fix: 处理左侧通宝解析失败的情况

travellerse:
## Summary by Sourcery

Handle cases where parsing the left copper column fails in roguelike exchange mode and tighten assumptions about detection results.

Bug Fixes:
- Return early with an error when the left copper column has zero or multiple detections instead of proceeding with invalid data.
- Fail fast if creating a new copper object from the left detection fails rather than silently continuing.

Enhancements:
- Simplify left-column handling to operate on a single expected detection and log clearer diagnostics, including saving a debug image in debug builds.

### DIFF
--- a/src/MaaCore/Task/Roguelike/JieGarden/RoguelikeCoppersTaskPlugin.cpp
+++ b/src/MaaCore/Task/Roguelike/JieGarden/RoguelikeCoppersTaskPlugin.cpp
@@ -560,7 +560,6 @@ std::optional<asst::RoguelikeCopper> asst::RoguelikeCoppersTaskPlugin::create_co
     return std::nullopt;
 }
 
-#ifdef ASST_DEBUG
 // 调试绘制辅助函数：在图像上绘制检测结果
 void asst::RoguelikeCoppersTaskPlugin::draw_detection_debug(
     cv::Mat& image,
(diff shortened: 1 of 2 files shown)


ABA2396:
save_debug_image得加个自动清理吧，感觉所有存图的地方都有个整个

ABA2396:
#14825

docs/zh-cn/protocol/integrated-strategy-schema.md
```diff
@@ -835,7 +835,7 @@ MAA 使用以下方法识别和处理通宝：
 
 ::: info 注意
 
-- 如果 OCR 识别的通宝名称在配置文件中未找到，MAA 会保存调试图像到 `debug/roguelikeCoppers/` 目录，以供排查
+- 如果 OCR 识别的通宝名称在配置文件中未找到，MAA 会保存调试图像到 `debug/roguelike/coppers/unknown` 目录，以供排查
```
ABA2396, line 838: i18n

ABA2396 (changes requested):

docs/zh-cn/protocol/integrated-strategy-schema.md
travellerse: 现在暂时还没有i18n，感觉到时候单开个PR写i18n比较好

PR #15359, opened 2026-01-10
fix: vertical alignment to top instead of center for updatesettings

Constrat:
The scenario:
in such a situation where I want to manually update MAA (when I want to) I always disable "auto download update".
The new update logging moves all the buttons on the left side making it impossible to re disable auto download update during an update process. 

This PR anchors everything to TOP, this way the user has a much cleaner experience to click buttons while keeping logging.

Something similar happens when enabling / disabling myrrorchyan, but on the right side buttons since the alignement is centered, the total height changes, shifting the buttons.

Screenshot 2026-01-10 150105
Screenshot 2026-01-10 150121


Constrat:
Are changes required? @ABA2396 it's still a very annoying issue, especially when manually updating MAA from a VNC via phone or similar.

PR #15754, opened 2026-02-20
feat: add abs_rect to ocr results

Constrat:
I'm not sure if this was ever wanted but the WordOcr `rect` result is always "wrong" or truncated?

Before: 
```py
asst::WordOcr [{ text: Credit Store, rect: [ 158, 98, 91, 18 ], score: 0.975565 }] by OCR Pipeline , cost 359 ms
Proceed [{ text: Credit Store, rect: [ 1158, 98, 91, 18 ], score: 0.975565 }]
```

After:
```py
asst::WordOcr [{ text: Credit Store, rect: [ 1158, 98, 91, 18 ], score: 0.975565 }] by OCR Pipeline , cost 321 ms
Proceed [{ text: Credit Store, rect: [ 1158, 98, 91, 18 ], score: 0.975565 }]
```

Not sure if this is the optimal solution? Maybe we can just work on the Logger itself?


status102:
Not sure it's a good idea. If input image of asst::WordOcr is not an entire image, adjustment would make no sense.

Constrat:
But currently it makes no sense. 
If there's is no ocrReplace being done, it will just show the wifth and height of the detected section. Making everything very complicated to debug. (Why was a character missing, is the size enough etc etc)

Credit Store, rect: [ 158, 98, 91, 18 ] is completely wrong... But it's very similar to the replaced result in the correct place. Is there just a logging issue? Like the x and y coords get truncated?

status102:
> Credit Store, rect: [ 158, 98, 91, 18 ] is completely wrong... But it's very similar to the replaced result in the correct place. Is there just a logging issue? Like the x and y coords get truncated?

It's not wrong. It's relative to the roi of `CreditStoreOcr`.

```
"CreditStoreOcr": {
    "algorithm": "OcrDetect",
    "text": ["信用交易所"],
    "roi": [1000, 0, 280, 300],
    "action": "ClickSelf",
    "next": ["CollectCredit", "CreditStoreFailed", "Stop"]
},
```

Constrat:
Oh. And what about OperBoxNameOCR?

The WordOcr result shows 0,0,70,19 (example for EN)

The 0,0 is based on the rect_move based on the flag? Because then that makes no sense, unless I'm missing a task.

(I'll try to get an example later)

The point of this PR is to simplify text detection when debugging. Having to do (simple) arithmetics just to check if a ROI is correct having to find the task is annoying.

status102:
Rect in output of OperBoxNameOCR makes no sense now. Because of the specific adjustment

Constrat:
It made no sense BEFORE.

Having it as 0,0 as anchor makes no sense because even if I "imagine" and calculate rect move, it still won't make the imaginary rectangle appear over the name.

Constrat:
image

Look at this.
To debug it I have to search for the GoChapterToGO task and ROI inside the stages navigation of official.

image

This is the ROI we expect and use got the task, but the log ROI makes no sense. when I compare it... 
This is what I'm annoyed about.


<hr>

Another instance:
image

Using the StageEncounterOcr ROI + the log one gets me this:
image


<hr>

All in all, it's a pretty annoying process to do calculations basing on the main task. The result even seems wrong at times? I don't understand what the issue is / if it's just a logging issue, then please point me in the right way.

status102:
At least, marking that rect has been adjusted please

Constrat:
I'm just going to add a third logging value then.

I'll leave rect as is and I'll try to think of a better name for the positionally correct rect

Constrat:
Yes it's all broken. I'm working on it. It will take more time.

I could just "steal" the data before the log.trace and edit it there by adding the absolute rect calculation but it looks like an ugly approach. I want to ideally add absolute rect to every OCR log

MistEO:
This is a bit strange. Why would we need an extra rect? Is it just for logging?

status102:
`asst::WordOcr [{ text: Credit Store, rect: [ 158, 98, 91, 18 ], score: 0.975565 }] by OCR Pipeline , cost 359 ms`
`asst::WordOcr [{ text: Credit Store, rect: [ 1158, 98, 91, 18 ], score: 0.975565 }] by OCR Pipeline and rect for full screen, cost 359 ms`

maybe some things like this? We just want to easy the simple pipeline task to check

Constrat:
> This is a bit strange. Why would we need an extra rect? Is it just for logging?

There is no immediate way of getting a ROI to test against a full sized screencap when checking for ocr issues. The current approach is to manually calculate the rect_move which as shown in the previous images doesn't even seem to align with the position of the text in the first place. 
In a lot of instances OCR results will display rect.x (and y) based on the roi task, instead of absolute coords.
 158, 98, 91, 18 vs  1158, 98, 91, 18
In this case it's simple, but in situations like OperBoxNaneOCR when there's a ton of results and the anchor flag moves it's a huge pain in the ass.

This would be useful not only as debug, but even in normal logging so that when users present their logs, it's immediately usable to check for ROI issues



It's not an "extra"rect. It's the coords of the real position in the screencap. Instead of a reference where we have to look for the anchor checking the task roi


> maybe some things like this? We just want to easy the simple pipeline task to check

Having multiple log lines for the "same" thing seems pretty polluting.. 
I'm trying to reach something like this.

`asst::WordOcr [{ text: Credit Store, abs: [ 1158, 98, 91, 18 ], rect: [ 158, 98, 91, 18 ], score: 0.975565 }] by OCR Pipeline , cost 359 ms`

status102:
> > maybe some things like this? We just want to easy the simple pipeline task to check
> 
> Having multiple log lines for the "same" thing seems pretty polluting.. I'm trying to reach something like this.

It's not multiple lines, but just one. 
For complex rect recognition performed by plugins, the corrected rect is often meaningless. However, for simple rect recognition performed by pipelines, the original rect is unnecessary once the corrected rect is available.

Constrat:
If the values are equal, there's no need for both. I've check the code better and it's based on the `without_det` flag. Whenever it's false, the displayed rect will not be an absolute coordinate, but a relative one

Constrat:
One more reasoning for this change (that I still have to think how to pull off without changing the entire codebase
Screenshot_2026-03-02-12-42-23-26_6844ed1968100643771cb3f60433f7e6

At a quick glance, this roi makes no sense without the corresponding task, which is usually not sent in logs extracts.

I think I'm just going to "dirty" the outpost of logs as it seems hard coded (the good looking text score rect roi section)

Constrat:
Ok. This took a bit but it's finally working as I expect it to.
```py
[2026-03-02 15:32:05.937][TRC][Px26852][Tx63884] asst::WordOcr [{ text: ture Store, abs_rect: [ 1000, 98, 78, 18 ], rect: [ 0, 98, 78, 18 ], score: 0.977025 }, { text: Credit Store, abs_rect: [ 1158, 98, 91, 18 ], rect: [ 158, 98, 91, 18 ], score: 0.975565 }] by OCR Pipeline , cost 314 ms
[2026-03-02 15:32:05.937][TRC][Px26852][Tx63884] Proceed [{ text: Credit Store, rect: [ 1158, 98, 91, 18 ], score: 0.975565 }]

```

src/MaaCore/Config/Miscellaneous/OcrPack.cpp
```diff
@@ -70,7 +70,7 @@ bool asst::OcrPack::load(const std::filesystem::path& path)
     return !m_det_model_path.empty() && !m_rec_model_path.empty() && !m_rec_label_path.empty();
 }
 
-asst::OcrPack::ResultsVec asst::OcrPack::recognize(const cv::Mat& image, bool without_det)
+asst::OcrPack::ResultsVec asst::OcrPack::recognize(const cv::Mat& image, bool without_det, const Rect& roi_offset)
```
status102:
```suggestion
asst::OcrPack::ResultsVec asst::OcrPack::recognize(const cv::Mat& image, bool without_det, const std::optional<Rect>& output_offset)
```

src/MaaCore/Config/Miscellaneous/OcrPack.cpp
```diff
@@ -121,13 +129,25 @@ asst::OcrPack::ResultsVec asst::OcrPack::recognize(const cv::Mat& image, bool wi
...
+        Rect screen_rect(roi_offset.x + abs_rect.x, roi_offset.y + abs_rect.y, det_rect.width, det_rect.height);
+        if (i > 0) {
+            raw_log += ", ";
+        }
+        raw_log += std::format(
+            "{{ text: {}, abs_rect: {}, rect: {}, score: {:.6f} }}",
+            ocr_result.text.at(i),
+            screen_rect.to_string(),
+            det_rect.to_string(),
+            ocr_result.rec_scores.at(i));
```
status102:
```suggestion
        if (i > 0) {
            raw_log += ", ";
        }
        if (output_offset) {
            Rect screen_rect(
                output_offset->x + abs_rect.x,
                output_offset->y + abs_rect.y,
                det_rect.width,
                det_rect.height);
            raw_log += std::format(
                "{{ text: {}, moved_rect: {}, rect: {}, score: {:.6f} }}",
                ocr_result.text.at(i),
                screen_rect.to_string(),
                det_rect.to_string(),
                ocr_result.rec_scores.at(i));
        }
        else {
            raw_log += std::format(
                "{{ text: {}, rect: {}, score: {:.6f} }}",
                ocr_result.text.at(i),
                det_rect.to_string(),
                ocr_result.rec_scores.at(i));
        }
```

src/MaaCore/Config/Miscellaneous/OcrPack.h
```diff
@@ -44,7 +44,7 @@ class OcrPack : public AbstractResource
 
     void use_gpu(int gpu_id) { m_gpu_id = gpu_id; }
 
-    ResultsVec recognize(const cv::Mat& image, bool without_det = false);
+    ResultsVec recognize(const cv::Mat& image, bool without_det = false, const Rect& roi_offset = {});
```
status102:
```suggestion
    ResultsVec recognize(const cv::Mat& image, bool without_det = false, const std::optional<Rect>& rect_offset = {});
```

status102 (commented):
The `abs_rect` in the output is the rect after offset, while the `abs_rect` in the code is the rect relative to the ROI. It's inconsistent.

Constrat:
Is rect_offset the correct naming tho?
It's not really the rect_move from the task. That's the base roi itself.

I think `base_roi` looks decent enough

status102:
> Is rect_offset the correct naming tho?

Oh, it's my fault. Called `output_offset` in function implement. Or is there other name you prefer?

Constrat:
No worries. Good catch on the std::optional. Tweaking the variables right now. I think it's ready for merge after that.

PR #15888, opened 2026-03-01
perf: 适配国服游戏内暂停期间部署

status102:
## Summary by Sourcery

增强功能：
- 更新部署流程，在支持“滑动并暂停”（swipe-with-pause）的环境中，在执行滑动部署之前触发显式暂停，从而简化滑动调用接口。

<details>

## Summary by Sourcery

Enhancements:
- Update deployment flow to trigger an explicit pause before performing swipe deployment in environments that support swipe-with-pause, simplifying the swipe call interface.

</details>

### DIFF
--- a/resource/global/YoStarEN/resource/tasks/tasks.json
+++ b/resource/global/YoStarEN/resource/tasks/tasks.json
@@ -1538,6 +1538,9 @@
         "rectMove": [0, 234, 116, 22],
         "specialParams": [105, 2, 7, 0, 3, 10]
     },
+    "BattleDeployWhenPause": {
+        "specialParams": [0]
+    },
     "BattleStartPre": {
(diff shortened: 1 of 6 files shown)


status102:
@Constrat 
This commit would break deploy with pause in oversea client.
Not sure is there some one would use this feature? Yj has made it is possible to deploy when pausing with the game's built-in pause

Constrat:
I mean can't we add a client check? I don't really like removing features like this. Even if the usage is not as popular, unless there's no way to differentiate at which point i guess global can go fuck itself

status102:
> I mean can't we add a client check? I don't really like removing features like this. Even if the usage is not as popular, unless there's no way to differentiate at which point i guess global can go fuck itself

fine. I would add a task to mark if new feature is enabled. And you may need to track it later

PR #15922, opened 2026-03-05
feat: 引入移除开机自启注册表项的脚本

lucienshawls:
## Summary by Sourcery

添加用于移除 Windows 自动启动注册表项的脚本，并将其包含在打包的安装工件中。

新功能：
- 提供一个批处理脚本，用于从 Windows 启动配置中移除自动启动注册表项。

构建：
- 在 CI 安装工件的打包过程中包含自动启动移除脚本。
- 在每夜构建的 OTA 发布打包过程中包含自动启动移除脚本。

<details>

## Summary by Sourcery

Add a script for removing Windows auto-start registry entries and include it in packaged installation artifacts.

New Features:
- Provide a batch script to remove auto-start registry entries from Windows startup configuration.

Build:
- Include the auto-start removal script in CI packaging of install artifacts.
- Include the auto-start removal script in nightly OTA release packaging.

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,6 +209,7 @@ jobs:
           rm -rf install/*.h
 
           cp tools/DependencySetup_依赖库安装.bat install
+          cp tools/RemoveAutoStartReg_移除开机自启注册表项.bat install
 
       - name: Zip files
         run: |
--- a/.github/workflows/release-nightly-ota.yml
+++ b/.github/workflows/release-nightly-ota.yml
@@ -263,6 +263,7 @@ jobs:
           rm -rf install/*.h
 
           cp tools/DependencySetup_依赖库安装.bat install
+          cp tools/RemoveAutoStartReg_移除开机自启注册表项.bat install
 
       - name: Upload MAA to GitHub
         uses: actions/upload-artifact@v7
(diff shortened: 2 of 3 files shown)


Constrat:
Ah I like the idea, but if someone completely deletes the MAA folder (with the .exe inside) then the .bat file will also get deleted, and the "shadow auto start" remains...

It's a bit of an unfixable issue... Sure they can re download MAA to run the .bat file and THEN remove completely MAA...

ABA2396:
It feels a bit redundant. When this batch file exists, it means MAA is still there, so you can just go into the settings and disable the auto-start option.

lucienshawls:
> Ah I like the idea, but if someone completely deletes the MAA folder (with the .exe inside) then the .bat file will also get deleted, and the "shadow auto start" remains...
> 
> It's a bit of an unfixable issue... Sure they can re download MAA to run the .bat file and THEN remove completely MAA...

To re-download and run this batch file does not necessarily guarantee the registry removal, unless the batch file is placed in the exact same directory as the `MAA.exe` file that was deleted earlier.



> It feels a bit redundant. When this batch file exists, it means MAA is still there, so you can just go into the settings and disable the auto-start option.

我最初是在想，运行一个 batch 文件要比进入 MAA 调设置要方便一点，因为其实难以保证用户真的能找到取消开机自启的设置（哪怕开机启动就是用户自己设置的）。

status102:
或许可以尝试和移除通知设置中的maa选项一起做成一个类似卸载的功能。 #8392

PR #15946, opened 2026-03-08
fix: EN RM SettlementStageName

Daydreamer114:
cc @Constrat 
Settlement Stage Name 是关卡结束后，结算界面的图片


Daydreamer114:
Where is the template optimization CI?

Constrat:
> Where is the template optimization CI?

I think precommit got killed. I really don't mind it xD

Iirc it automatically runs weekly or something

PR #15758, opened 2026-02-21
ci: updates to Resource Updater and clangd

Constrat:
Runs fine.
- https://github.com/MaaAssistantArknights/MaaAssistantArknights/actions/runs/22246692939/job/64362113692

## Summary by Sourcery

简化 ResourceUpdater 在物品分类和基础设施房间类型方面的数据处理，并优化 CI 工作流，在使用缓存二进制文件时避免不必要的子模块拉取。

Bug Fixes:
- 通过移除旧的数字到字符串映射，使物品分类和基础设施房间类型的处理与当前游戏数据格式保持一致。

Enhancements:
- 假定 `classifyType` 和 `roomType` 字段为字符串类型，以简化物品和基础设施的 JSON 转换。

CI:
- 更新 `res-update-game` 工作流，使其仅在 ResourceUpdater 缓存未命中时才拉取 MaaUtils 子模块。

<details>
<summary>Original summary in English</summary>

## Summary by Sourcery

Simplify ResourceUpdater data handling for item classification and infrastructure room types, and optimize the CI workflow to avoid unnecessary submodule fetches when using cached binaries.

Bug Fixes:
- Align item classification and infrastructure room type processing with the current game data formats by removing legacy numeric-to-string mappings.

Enhancements:
- Simplify item and infrastructure JSON transformations by assuming string-based classifyType and roomType fields.

CI:
- Update the res-update-game workflow to only fetch MaaUtils submodule when the ResourceUpdater cache is missed.

</details>

PR #16116, opened 2026-03-27
ci: 避免生成不必要的res_updater

status102:
res-update-game.yml 中使用ResourceUpdater的自己的cmakelist，上层中似乎不应启用res_updater编译

@soundofautumn 

## Summary by Sourcery

更新 CMake 预设，以避免在 CI 中构建不必要的 `res_updater` 目标，并与专用的 ResourceUpdater CMake 配置保持一致。

Build:
- 调整 CMakePresets 配置，在不需要的情况下禁用或省略 `res_updater` 的构建。

CI:
- 优化 CI 配置，使 `res-update-game` 工作流依赖 ResourceUpdater 自身的 CMakeLists，而不是构建顶层的 `res_updater` 目标。

<details>

## Summary by Sourcery

Update CMake presets to avoid building the unnecessary res_updater target in CI and align with the dedicated ResourceUpdater CMake configuration.

Build:
- Adjust CMakePresets configuration to disable or omit the res_updater build where it is not required.

CI:
- Refine CI configuration so res-update-game workflow relies on ResourceUpdater’s own CMakeLists instead of building the top-level res_updater target.

</details>

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -18,6 +18,7 @@
             "cacheVariables": {
                 "BUILD_WPF_GUI": "ON",
                 "BUILD_DEBUG_DEMO": "ON",
+                "BUILD_RESOURCE_UPDATER": "ON",
                 "INSTALL_RESOURCE": "OFF",
                 "INSTALL_PYTHON": "OFF"
             },
@@ -170,7 +171,8 @@
                 "INSTALL_RESOURCE": "ON",
                 "INSTALL_PYTHON": "ON",
                 "BUILD_WPF_GUI": "OFF",
-                "BUILD_DEBUG_DEMO": "OFF"
+                "BUILD_DEBUG_DEMO": "OFF",
+                "BUILD_RESOURCE_UPDATER": "OFF"
             }
         },
         {
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -18,7 +18,6 @@
             "cacheVariables": {
                 "BUILD_WPF_GUI": "ON",
                 "BUILD_DEBUG_DEMO": "ON",
-                "BUILD_RESOURCE_UPDATER": "ON",
                 "INSTALL_RESOURCE": "OFF",
                 "INSTALL_PYTHON": "OFF"
             },
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -178,8 +178,8 @@
         {
             "name": "windows-publish-x64",
             "inherits": [
-                "windows-x64",
-                "publish-base"
+                "publish-base",
+                "windows-x64"
             ],
             "$comment": [
                 "github actions only support Visual Studio 17 2022",
@@ -191,8 +191,8 @@
         {
             "name": "windows-publish-arm64",
             "inherits": [
-                "windows-arm64",
-                "publish-base"
+                "publish-base",
+                "windows-arm64"
             ],
             "$comment": [
                 "github actions only support Visual Studio 17 2022",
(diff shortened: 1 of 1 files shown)


soundofautumn:
#15758 似乎是 @Constrat 引入的这个res_updater

Constrat:
image
The issue is this. When working on the main project window res updater build in the cmake extension will disappear. 

Can't work on resource updater through cmake extension when on the main project dir. I did not notice it would ship out the `res_updater.exe` to prod and downloads. 
(It's just 600kb and without gamedata is does nothing...)

We can just `rm` it in workflow runs? It's not the most elegant solution but I couldn't really anything else to add the resource updater as a target from the main CMake file. And let's not "open resource updater in its own environment" because that's just annoying and not really a solution

status102:
> image The issue is this. When working on the main project window res updater build in the cmake extension will disappear.
> Can't work on resource updater through cmake extension when on the main project dir. I did not notice it would ship out the `res_updater.exe` to prod and downloads. (It's just 600kb and without gamedata is does nothing...)
> 
> We can just `rm` it in workflow runs? It's not the most elegant solution but I couldn't really anything else to add the resource updater as a target from the main CMake file. And let's not "open resource updater in its own environment" because that's just annoying and not really a solution

```
if(DEFINED ENV{VSCODE_PID})
        add_custom_target(run-MaaWpfGui
            COMMAND "${CMAKE_BINARY_DIR}/bin/$<CONFIG>/MAA.exe"
            WORKING_DIRECTORY "${CMAKE_BINARY_DIR}/bin/$<CONFIG>"
        )
        add_dependencies(run-MaaWpfGui MaaCore MaaWpfGui)
    endif()
```

I use this to resolve the problem that `run-MaaWpfGui` would be shown in VS. Not sure if this could solve that

soundofautumn:
maybe we can simply add                "BUILD_RESOURCE_UPDATER": "OFF", in publish-base, and the inherit order should also change from publish-base, windows-x64 to windows-x64, publish-base

status102:
> maybe we can simply add "BUILD_RESOURCE_UPDATER": "OFF", in publish-base, and the inherit order should also change from publish-base, windows-x64 to windows-x64, publish-base

Not only publish, maybe we could also remove from VS resolution manager. Seems no one use it in VS

soundofautumn:
> > maybe we can simply add "BUILD_RESOURCE_UPDATER": "OFF", in publish-base, and the inherit order should also change from publish-base, windows-x64 to windows-x64, publish-base
> 
> Not only publish, maybe we could also remove from VS resolution manager. Seems no one use it in VS

then nobody use it in Vscode either, I think it's not the problem with VS and Vscode like run-MaaWpfGui. It seem the res_updater is rarely need to develop or build.

Constrat:
...I use both of them and it works? Just because you don't use it doesn't mean someone else doesn't as well... The reason I added res_updater was because I used it.

Regarding run-MaaWpfGui I'm not sure what you're talking about, I just changed it so that it also builds the MaaCore, that way you can refresh and still build the core every time cmake uses chaching anyway


<hr>

If the resource updater requires a fix or even more development I'm not starting VS26, especially when I also have to check the resources and run the gamedata scripts.

status102:
Fine. So we just need to remove it from publish

Constrat:
> Fine. So we just need to remove it from publish

This should work yes?

cc @soundofautumn

soundofautumn:
> > Fine. So we just need to remove it from publish
> 
> This should work yes?
> 
> cc @soundofautumn 

the inherit order also should change so that the publish-base variable can take effect

```
             "name": "windows-publish-x64", 
              "inherits": [ 
                  "publish-base", 
                  "windows-x64" 
              ],
```

```
             "name": "windows-publish-x64", 
              "inherits": [ 
                  "windows-x64",
                  "publish-base"
              ],
```

Constrat:
@soundofautumn I made the commit with the change but...
image

This is the run from the previous commit ([068fd89](https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/16116/commits/068fd89635011c2bdca401a1a197652b251c489c)) and as you can see the flag BUILD_RESOURCE_UPDATER is OFF anyway, even without moving them?

NOW with the change [e85b8e4](https://github.com/MaaAssistantArknights/MaaAssistantArknights/pull/16116/commits/e85b8e4ad5155a71bbb232a1ca069eba91907909) the flag is ON 
image

<hr>

> If multiple inherits presets provide conflicting values for the same field, the earlier preset in the inherits list will be preferred

It's priority based, first come first serve (not really but you get the idea), unlike our tasks.json where it follows all the steps from top to bottom applying the bottom one last

PR #16937, opened 2026-05-30
Revert "fix: 调整基建干员名 OCR 区域"

ZiyinLin:
Reverts MaaAssistantArknights/MaaAssistantArknights#16905

## Summary by Sourcery

改进：
- 在 `tasks.json` 中恢复此前用于基础设施运营商名称检测的 OCR 配置。

<details>

## Summary by Sourcery

Enhancements:
- Restore the prior OCR configuration for infrastructure operator name detection in tasks.json.

</details>

### DIFF
--- a/resource/global/YoStarEN/resource/tasks/tasks.json
+++ b/resource/global/YoStarEN/resource/tasks/tasks.json
@@ -1259,9 +1259,6 @@
     "ShamareThumbnail3": {
         "maskRange": [1, 255]
     },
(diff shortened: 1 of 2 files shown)


Roland125:
@Constrat  After conducting more tests, we found that this OCR range modification also has certain negative effects on the Simplified/Traditional servers, so we believe it is necessary to revert this change. 

Please confirm that your manual settings on the EN server are the same as the OCR range parameters after this revert.

PR #17516, opened 2026-08-02
fix: prevent OCR replace chain from recognizing TA characters repeatedly

Ronnoc:
In CharsNameOcrReplace, rules are applied as a sequential chain: ["エフイーター", "食铁兽"] correctly translates the JP OCR text for FEater, but the later rule ["^「?食[」\\]]?", "食字"] (intended for the single-kanji clue/item named 食) lacks an end anchor, so it re-matches the already-replaced "食铁兽" string (since it only needs to start with 食) and overwrites it with "食字".

As a result, FEater is silently unrecognized during infrast custom-operator selection on the JP client, even though the initial OCR read (エフイーター) was correct.

Add a $ end anchor to the 食 single-character rule so it only matches the standalone token, not any text that merely starts with 食.

## Summary by Sourcery

错误修复：
- 通过阻止仅匹配“食”的规则重新匹配并覆盖，确保在日文 OCR 名称替换过程中能够正确识别 FEater（食铁兽）。

<details>

## Summary by Sourcery

Bug Fixes:
- Ensure FEater (食铁兽) remains correctly recognized during JP OCR name replacement by preventing the 食-only rule from rematching and overwriting it.

</details>


status102:
@Manicsteiner 
- link to #17389

Manicsteiner:
That's a good suggestion, but in actual OCR recognition, there might be a string of meaningless characters following normal content, so I wouldn't really recommend modifying it that way. I'll try to make a better fix, like `["^(「食|食[」\\]]|食$)", "食字"]`

PR #17945, opened 2026-08-29
feat(pc): PC端连接方式默认隐藏，仅可通过配置 AllowPCClient 启用

ocsin1:
### 改动

- 新增根配置 `Gui.AllowPCClient`（位于 `gui.new.json` 的 `Gui` 节点，默认 `false`）
- 未启用时，连接设置下拉框过滤掉 `ConnectConfig.PC` 项（连接设置 / 启动设置 / 新手引导三处入口）
- 未启用时，存量已选用 PC 连接的配置启动时静默回退到 `General`（通用连接），保留 ADB 路径/地址与 Win32 附加设置，便于开启后恢复

### 测试

- [x] 默认隐藏：未启用时下拉框无 PC 项
- [x] 存量回退：`Config = "PC"` 且未启用时，启动不崩溃，配置自动回退 `General`，ADB/Win32 设置保留
- [x] 配置启用：`AllowPCClient = true` 时 PC 项出现且不被回退

## Sourcery 摘要

默认隐藏 PC 客户端连接方式，并允许通过配置启用。

新功能：
- 添加 `Gui.AllowPCClient` 设置，用于控制 PC 客户端连接选项的可见性。

错误修复：
- 默认隐藏 PC 客户端连接选项，并自动将现有的 PC 配置回退为“通用”，同时保留相关连接设置。

增强功能：
- 在连接设置中统一应用 PC 连接筛选，并安全处理启动时的回退，避免初始化错误。

<details>

## Summary by Sourcery

Hide the PC client connection method by default and allow it to be enabled through configuration.

New Features:
- Add the Gui.AllowPCClient setting to control visibility of the PC client connection option.

Bug Fixes:
- Hide the PC client connection option by default and automatically fall back existing PC configurations to General while preserving related connection settings.

Enhancements:
- Apply PC connection filtering consistently across connection settings and safely handle startup fallback without initialization errors.

</details>

### DIFF
--- a/src/MaaWpfGui/Configuration/Global/Gui.cs
+++ b/src/MaaWpfGui/Configuration/Global/Gui.cs
@@ -45,6 +45,12 @@ public class Gui : NotifyPropertyChangedWithValue
 
     public bool HideCloseButton { get; set; }
 
+    /// <summary>
+    /// 是否在连接设置中显示"明日方舟 PC 端（社区维护）"连接方式。
+    /// 默认隐藏，仅可通过修改配置文件（gui.new.json 的 Gui 节点）启用。
+    /// </summary>
+    public bool AllowPCClient { get; set; }
+
     public string WindowTitleSelectShowList { get; set; } = "2 3 4";
 
     public bool WindowTitleScrollable { get; set; }
--- a/src/MaaWpfGui/ViewModels/UI/SettingsViewModel.cs
+++ b/src/MaaWpfGui/ViewModels/UI/SettingsViewModel.cs
@@ -331,6 +331,26 @@ public class SettingsViewModel : Screen
         ConnectSettings.ConnectAddressHistory.CollectionChanged += (_, _) => {
             ConfigFactory.CurrentConfig.Gui.ConnectSettings.AddressHistory = [.. ConnectSettings.ConnectAddressHistory];
         };
+
+        // 明日方舟 PC 端连接方式默认隐藏（AllowPCClient 默认 false，仅改配置文件启用）。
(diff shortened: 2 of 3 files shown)


H2O-MERO:
？为什么 反对

Constrat:
@ocsin1 This is too late. (kind of) we may have to look for different solutions (more red text)

@H2O-MERO regarding your question. I'm sure you're aware of the differences (and inconsistencies) between the emulated (android) and PC client, from the cursor being occasionally in the way text rendering, template size etc etc will lead to hard to debug things. So ideally only knowleadgeable / power users will enable this option, that way instead of the usual useless issue "it's not working" we may be able to create a small conversation and perhaps even let the user fix it by themselves and opening a PR. As mentioned basically everywhere MAA was created when there was only the "emulator" client. The PC client adaptation is, at-best a huge patch that will sooner or later break. You (or users) may say "but everything works for me" but it's the same situation with forcing the 1080p resolution for YostarEN. Everything _may_ work. It's a very different thing as we don't guarantee anything and guaranteeing a 100% working pc client version still requires A LOT of work.

H2O-MERO:
@Constrat There are special optimizations for the PC client in the released test builds. I would like to put this PR on hold before the official release and observe user feedback for some time.

Constrat:
Absolutely, that's why I said a PR like this "is late". This was supposed to be released together with the first early implementation of pc support. Now the ship has already sailed.

ABA2396:
现在在需要高频截图的场景，基本上已经鼠标锁在左下角了，比如自动战斗或者肉鸽，甚至部分识别较快的牛杂

PR #18273, opened 2026-09-19
docs: 更新readme中的截图资源，并对圆角外使用透明通道（zh-cn）

H2O-MERO:
界面已经改版过了，开个pr更新一下资源，因为我没有牛牛所以干员识别更新不了...


H2O-MERO:
简中之外的语言，因为我只有国服，所以截图效果不好，只做了zh-tw的

H2O-MERO:
@ABA2396 先只对简中readme进行更新？

Constrat:
This is becoming quite tiresome to do every few months. As trying to get the "most beautiful" interface with all the settings while trying to keep it consistent with the other screenshots is basically impossible.

I understand wanting to keep the images up to date with the latest MAA features but still. 
I'm going to be straight in saying: I really don't have a solution to propose compared to the current way of doing things such which is getting some debug tasks / cool recruitment stuff to get the most complete MAA showcase experience.


Also, on another note, I don't suggest using the PC client for these images, as PC client support remains "at your own risk / may not work). Publicly showcasing it like that is not ideal.

H2O-MERO:
@Constrat There is no requirement for a unified content standard across interfaces in different languages. After all, users will only look at screenshots in their own language and won't compare them, so it's enough to take a suitable screenshot for each language's interface—there's no need to make it "most beautiful."

As for the PC screenshot issue, I had considered this from the start. Since you brought it up, I think we can go ahead and make changes.

Constrat:
@H2O-MERO But I'm a bit autistic for that sort of things... So if a language shows the "goodies" I feel like other languages should show them as well xD.

My biggest complaint remains that it would be pretty cool if we could somehow automate this (to get it standardized to the best possible) but I'm pretty sure github runners are limited when trying to pull this stuf..

H2O-MERO:
@Constrat  Perhaps we could export the "goodies" config files for unification? For servers in different regions, I don't have a good solution either...this is all I can think of. 

As for the output logs, if auto recruitment is meant to be showcased, then maybe there's no need to unify the recruitment logs; just start the task before taking a screenshot :)

H2O-MERO:
@Constrat Then maybe we could automate local screenshots? I did write a program to automatically take screenshots and save them as files🤔

Constrat:
I like the config thingy. For consistency! Obviously it would need to have a different client for each. But having a separate config for screenshots always available (where one would just need to get the recruitment section) as it's definitely the "easiest" and fastest to get.

Since everything else is just operator detection and depot detection.

H2O-MERO:
@Constrat  yep, the last three screenshots are all very easy to unify.

Constrat:
Btw, your rounded screenshots, have background on the right hand side... Is that wanted?
Being on win 11, I use a simple .exe that removes every rounded corner so I just disable it to get the rounded, then I manually go into photoshop to add transparency to the corners as i take the screens with a full black background so it's easier to "remove"

H2O-MERO:
And how should the standards be set for operator and depot , as well as for the task list?

Considering different languages, the task list cannot be renamed and must use the default

Constrat:
> And how should the standards be set for operator and depot , as well as for the task list?

~You mean the export? You can choose whatever, I don't think  there's a "best" solution~
image

Well I'd say use MAA's default names.
Perhaps hovering OVER the task to show you can rename / delete / copy etc
image

H2O-MERO:
> Btw, your rounded screenshots, have background on the right hand side... Is that wanted? Being on win 11, I use a simple .exe that removes every rounded corner so I just disable it to get the rounded, then I manually go into photoshop to add transparency to the corners as i take the screens with a full black background so it's easier to "remove"

Since I also use Win11, taking a direct screenshot would capture the content behind the window as well, so I used a program to round the corners.

Constrat:
https://github.com/rich-ayr/win11-toggle-rounded-corners
If you hate rounded corners I highly suggest ^
Has been working wonderfully for me

But I guess in a website images they do look better... Also. How do you guys get 800x600 screenshots? Is that because of 1080p?

I can only get 1000x750. Do I manually scale it down?

H2O-MERO:
> You mean the export? You can choose whatever, I don't think there's a "best" solution

For example, do we need to show that we have all operators, and do we need them all at max level?

Constrat:
> For example, do we need to show that we have all operators, and do we need them all at max level?

Oh no... Definitely not, that would be min maxing A BIT TOO MUCH. Just get the usual maintainer account is fine (I think).
Same thing for the depot. Just get how you can.

IDEALLY, since that's NOT log dependant... A single maintainer (AK account) could do one for all clients tho.

For example:
image
I can change MAA's language and we could have my operators for every screenshot.
I would have to change the top line so that it follows the "configs" mumu / pc client / arknights client / event etc etc.
But that just requires me to have the configs of all clients.

H2O-MERO:
> Perhaps hovering OVER the task to show you can rename / delete / copy etc

great i like this :)

H2O-MERO:
> https://github.com/rich-ayr/win11-toggle-rounded-corners If you hate rounded corners I highly suggest ^ Has been working wonderfully for me

wow thank you

H2O-MERO:
> But I guess in a website images they do look better... Also. How do you guys get 800x600 screenshots? Is that because of 1080p?

On my computer, when I zoom out all the way, it's 800x600, which matches the original image size. But I feel like it's a bit small, making the elements look a bit cramped

H2O-MERO:
> For example:

This example looks pretty good. I've never had Pallas, which forced me to manually edit the JSON file XD

Constrat:
Wait! We can just standardize it in the config itself. The owned / not owned ops are in the config json file.
So a single configs (which is then "adapted" for the 5 clients) could actually work. And one maintainer could get ALL of them

H2O-MERO:
> I would have to change the top line so that it follows the "configs" mumu / pc client / arknights client / event etc etc.

Does the title bar also need to be unified...? I don't think it's really necessary...🤔

Constrat:
> Does the title bar also need to be unified...? I don't think it's really necessary...🤔

It does inside the same client. Ideally... It would be cool having it standardized where just the event / resources / client changes.

H2O-MERO:
> Wait! We can just standardize it in the config itself. The owned / not owned ops are in the config json file.

Yes, great idea 🤓

But the thing is, a maintainer won't be managing server clients for different regions at the same time. The server region in the title bar should still match the interface language, right?🤔

Constrat:
See how my images are bigger lol :O

H2O-MERO:
huge image but looks more comfortable XD

btw, the interface size also need to be unified...?

ABA2396:
600\*800 是标准大小，Windows 缩放 125% 截图截出来就是 750\*1000

H2O-MERO:
@ABA2396 虽然600 800好控制，但是感觉太挤了

ABA2396:
这玩意的缩略图应该看不清的吧，直接切换界面语言截图不就好了吗？我看看能不能整个预制的出来

H2O-MERO:
> 这玩意的缩略图应该看不清的吧，直接切换界面语言截图不就好了吗？我看看能不能整个预制的出来

是一个需不需要用对用语言服务器客户端的问题

ABA2396:
#18323 自动截图超集覆盖

